PR #26, opened 2015-04-23
Implement "running status" when saving MidiFiles.

mrjbq7:
I believe this implements the "running status" support for saving MidiFiles.

Before this commit, I had a MIDI file that was 37kb become 49kb when saved.  After this commit, it is the same size (well actually, 8 bytes smaller).

Be good to have a test case, but I don't see any MIDI files checked into the repository to use as a test case?

### DIFF
--- a/mido/midifiles.py
+++ b/mido/midifiles.py
@@ -446,6 +446,7 @@ class MidiFile:
 
             for track in self.tracks:
                 bytes = []
+                running_status_byte = None
                 for message in track:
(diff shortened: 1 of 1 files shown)


olemb:
Wow, thanks! I've been meaning to add this but I haven't gotten around to it.

I've tested a bit and it handles all edge cases I could think of, so I've merged your changes into the develop branch.

Yes, some test files would be good and also unit tests for MidiFile.

PR #29, opened 2015-08-16
Added the ability to read and write file objects as well as filenames

Twangist:
Added the ability to use file objects as well as filenames when reading, writing and saving MIDI files. This allows you to create a MIDI file dynamically (possibly *not* using `mido`), save it to an `io.BytesIO`, and then play that in-memory file, without having to create an intermediate external file. Of course the file (and/or the MidiFile) can still be saved to an external file.

midifiles.py: changes to ByteReader, ByteWriter, MidiFile.save. Parameters and instance attribute `filename` renamed to `filename_or_file`.
test_midifiles2.py: unittests of new feature
messages.py: corrected docstrings, error messages
__init__.py: version number bump

Docs: midi_files.rst -- corrected an example; added a paragraph (which could be better-placed) on new feature

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,10 +4,20 @@ Changes
 Release History
 ---------------
 
+1.1.15 (2015-08-16)
+^^^^^^^^^^^^^^^^^^^
+* Added the ability to use file objects as well as filenames when reading,
+  writing and saving MIDI files. This allows you to create a MIDI file
+  dynamically, save it to an io.BytesIO, and then play that in-memory
+  file, without having to create an intermediate external file. Of course the
+  file (and/or the MidiFile) can still be saved to an external file.
+
+  (Implemented by Brian O'Neill.)
+
 1.1.14 (2015-06-09)
 ^^^^^^^^^^^^^^^^^^^
 
-* bugfix: merge_tracks() concatinated the tracks instead of merging
+* bugfix: merge_tracks() concatenated the tracks instead of merging
   them.  This caused tracks to be played back one by one. (Issue #28,
   reported by Charles Gillingham.)
 
--- a/docs/midi_files.rst
+++ b/docs/midi_files.rst
@@ -2,7 +2,7 @@ MIDI Files
 ==========
 
 MidiFile objects can be used to read, write and play back MIDI
-files. (Writing is not yet implemented.)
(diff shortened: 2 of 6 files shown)


Twangist:
On further reflection, I'm not that fond of changing parameter names from `filename` to `filename_or_file`. It could break someone's code if they actually used the keyword `filename`. A different, probably better, approach: add *another* keyword parameter: `file=None'.
If you prefer that, as I now do, *don't* accept my pull request yet! I'll make the change and push it to the fork's master. 
My compliments on a very nice library, by the way. I like your Python very much :)

Twangist:
(Did that.)

olemb:
That's a great idea. I also appreciate that you've fixed a number for typos and a very broken example. 

Unfortunately I accidentally merged to my master branch, which is supposed to contain the latest release, so I had to revert the merge. I am trying to figure out how to revert the revert in the develop branch. What a mess.

I have some comments and ideas. I'll get back to you on that once I've figured out how to clean up the repository.

Thanks for your contribution!

Twangist:
Hi Ole,
Thank you! & you're very welcome.

I chuckled when I saw you accidentally merge into the Master branch, revert, then merge into dev branch — something I would do :)  It's funny how the last, "easiest", almost trivial step of a development cycle can be so easy to not get right the first time. Probably after all the careful work we let down our guard: "Ok, now *this* will be a breeze"… and it almost never is.
I'm sure you'll get the repo into shape, maybe you already have.

And yes, please do keep in touch & let me know your further thoughts.

Regards,
— Brian

olemb:
So, that took a while.

It turns out all I had to do was "git revert" on the commit that reverted you changes.

I've refactored the code a bit to make it simpler. ByteReader and ByteWriter now always take a file. It's up to the MidiFile object to open the file.

I also removed the file_or_filename attribute as it is no longer needed. (The file object is not kept around after the data has been read.)

The track comparisons in mido/test_midifiles2.txt should probably compare tracks instead of strings. Unfortunately since I chose to skip the time attribute in message comparisons you can't just do ``track1 != track2``. Something like this should work though:

    def compare_tracks(track1, track2):
        return [(msg, msg.time) for msg in track1] == [(msg, msg.time) in track2]

PR #31, opened 2015-08-27
Added descriptive Host Error messages to PortMidi backend

TRManderson:
In the case of a PortMidi host error, uses PmGetHostErrorText, instead of just PmGetErrorText. There might be a better way to deal with functions asking for buffers using ctypes, but I need a solution quickly so I just did it straight up.


olemb:
Thanks! This will make error messages a lot more helpful.

I have merged your changes to my develop branch and refactored it a bit. I also upped the max length from 80 to 256 since that's the value used in portmidi.h.

PR #38, opened 2016-03-22
use the correct __init__ for PortServer

njhurst:
I couldn't get PortServer to run (missing field yield_ports) because it was using the wrong base class `__init__`.


olemb:
Thanks!

I've merged it into my develop branch.

I wonder why it worked for me before. What happened when you tried to run it?

njhurst:
It crashed here: https://github.com/olemb/mido/blob/master/mido/ports.py#L325

PR #60, opened 2016-09-25
Fix sending long sysex messages with rtmidi backend

dagargo:
There is a problem (ALSA: seq_midi: MIDI output buffer overrun) when sending very large sysex messages. However, I found a solution and a workaround when working with the rtmidi backend.

The solution is based on [a python-rtmidi example](https://github.com/SpotlightKid/python-rtmidi/blob/master/examples/sendsysex.py). Basically, if the message is too large it must be splitted and the application must wait for the buffer to be empty before sending more data, otherwise the MIDI output buffer overrun occurs.

### DIFF
--- a/mido/backends/rtmidi.py
+++ b/mido/backends/rtmidi.py
@@ -9,6 +9,9 @@ import time
 import rtmidi
 from ..ports import BaseInput, BaseOutput
 
+MSG_LEN = 64
+SLEEP_TIME = 0.02
+
 def _get_api_lookup():
     api_to_name = {}
     name_to_api = {}
@@ -138,4 +141,12 @@ class Input(PortCommon, BaseInput):
 
 class Output(PortCommon, BaseOutput):
     def _send(self, message):
-        self._rt.send_message(message.bytes())
+        if message.type == 'sysex':
+            t = message.bytes()
+            while len(t) > MSG_LEN:
+                h, t = t[:MSG_LEN], t[MSG_LEN:]
+                self._rt.send_message(h)
+                time.sleep(SLEEP_TIME)
+            self._rt.send_message(t)
+        else:
+            self._rt.send_message(message.bytes())


olemb:
Hi,

Thanks for the patch! This has come up before, so it's good to get an implementation.

I'd like to include this but I'm not sure how best to go about it.

The change will introduce a timeout even for users who are not affected by this problem. Perhaps something like this could help narrow it down:

    if msg.type == 'sysex' and self._device_type == 'RtMidi/LINUX_ALSA':

although it would still affect all ALSA users.

I would like to find out what's actually happening here. I've been testing with a couple of devices and here it seems to max out at a bit over 16000 bytes (17000 for example), but with a different error message than yours:

    MidiOutAlsa::sendMessage: error sending MIDI message to port.

Here's the program I used:

```
import sys
import mido

mido.set_backend('mido.backends.rtmidi')

def send_sysex(port, size):
    print('Sending {} bytes to {}...'.format(size, port.name))
    sys.stdout.flush()
    port.send(mido.Message('sysex', data=[0] * size))

port = mido.open_output()

send_sysex(port, 40)
send_sysex(port, 100)
send_sysex(port, 16000)
send_sysex(port, 17000)
```

I'm curious about your max value of 64 bytes. Are you getting the error with messages this small? If so maybe there is a limit in the driver. May I ask you which device you're sending to? Could you run my test program and send me the output?

There are some small style issues with your code, but they are easily fixed and we can get back to them later.

Sorry if this is a lot of questions. I just want to get this right. :)

olemb:
I just realized you're the same person who reported this problem earlier. I'm glad you found a workaround. I hope we can figure out how to add this to Mido so you won't have to keep patching it.

dagargo:
I want to get this right too, so don't bother about asking too many questions.

First of all, I just want to clarify that the numbers I chose for MSG_LEN and SLEEP_TIME were arbitrary. However, the ratio between them was not, as it is an approximation of the MIDI serial bandwidth, which is 31250b/s or 3125B/s.

I run your program in 3 devices for a total of 5 midi ports and the output was the same you got.
```
Sending 40 bytes to Midi Through 14:0...
Sending 100 bytes to Midi Through 14:0...
Sending 16000 bytes to Midi Through 14:0...
Sending 17000 bytes to Midi Through 14:0...

MidiOutAlsa::sendMessage: error sending MIDI message to port.
```

Later, I run some tests to see which was the best ratio for a Moog LP II, which was connected thru a USB port.
For MSG_LEN up to 4kB, the time depends on the number of messages but I can conclude that the maximum bandwith the synth accepts is between 3200B/s and 8kB/s. For MSG_LEN grater than 4kB it does not work, so sending the message to the device does not mean the device can process the message right. Also, it worths mentioning that in this case the values I proposed were not optimal.
```
MSG_LEN;SLEEP_TIME;ok
16384;5;no
8k:5;no
8k:1;no
8k;0.5;no
4k;0.5;yes
4k;0.3;no
2k;0.25;yes
2k;0.2;no
1k;0.125;yes
1k;0.1;no
512;0.1;yes
512;0.05;no
256;0.05;yes
256;0.025;no
128;0.025;yes
128;0.0125;no
64;0.0125;yes
64;0.005;no
```
So It looks like there are other things to be considered.
I have been investigating and many of the libraries and utilities used to transfer sysex files use a configurable delay which is said to depend on the synth and computer configuration. Maybe, factors as the synth CPU, its flash memory and, of course, the program that runs in it play an important role. All of this make sense.

However, with amidi there is no need to set any delay and the sysex transmission is very fast.

I will keep investigating this because I would like to know what is really happening here.

olemb:
Ah, I see. There are a lot of factors to consider here. I'm curious how
amidi handles this.

dagargo:
I may have found the reason for the differences between the tools. Whereas [amidi](http://git.alsa-project.org/?p=alsa-utils.git;a=blob;f=amidi/amidi.c;h=a8264f181cf3a89c26ef3451815a18bb6a17f41a;hb=HEAD) uses the [rawmidi interface](http://www.alsa-project.org/alsa-doc/alsa-lib/rawmidi.html), [RtMidi](https://github.com/thestk/rtmidi/blob/master/RtMidi.cpp) uses the [sequencer interface](http://www.alsa-project.org/alsa-doc/alsa-lib/seq.html). In case of sending a large sysex message we should be using the rawmidi interface, which is a lower level interface that does not consider timing. The sequencer interface is focused on sending small messages in 'real-time' and considers timing.

Regarding our issue, I read in a [forum](http://article.gmane.org/gmane.linux.audio.devel/18617) that in order to send large sysex messages thru the sequencer interface they should be splitted in chunks so every chunk would be "scheduled at proper times so that the MIDI bandwidth is not exceeded" and maybe this is the reason for the delay between chunks.

Finally, although not related to our issue, I would like to add that a [commit](http://git.alsa-project.org/?p=alsa-utils.git;a=commit;h=b59745978fe7071e716234d80d715bc87822ff89) has recently been added to amidi in order to be able to sleep between different sysex messages to give the devices time to process them. This is what I mentioned in my previous comment and it is a different thing than adding a delay in between chunks of the same message.

I hope all this clarifies our misunderstanding of the issue.

olemb:
I have considered adding a rawmidi backend.

Thanks for the clarification.

Sleep between messages can easily be added in the application, while sleep inside a message needs to be added to the backend. How about something like this?

```
mido.open_output(sysex_interval=0.001)
```

where the default would be to not sleep.

dagargo:
I think that the rawmidi backend is our best option. How are you planning to do so? Maybe, I could help with it.

Regarding your proposal, it looks like a very nice addition but splitting still has to be considered by the Mido API users. So maybe it is better to focus on the new backend and solve the real problem.

And thanks for your support.

olemb:
I agree. The API should be the same for all backends. Also, let's start with RawMidi as you suggest.

I appreciate any help I can get. :)

I've opened a new issue where we can discuss the RawMidi backend.

olemb:
I think you are subscribed to the new issue since I mentioned you there, but in case not it's issue #66.

olemb:
I've added a very experimental amidi backend: https://github.com/olemb/mido/blob/develop/mido/backends/amidi.py

The latency is way too high for realtime use but maybe it's OK for sysex dumps. I'd like to find a way to pipe things through amidi though so we don't have to run it once for each message.

dagargo:
Hi!
I don't think this PR is useful any more. Should we close it?
Thanks.

olemb:
OK with me. We can always reopen it or create a new one if needed.

Issue #63, opened 2016-10-13
IOError('data byte must be in range 0..127') with Lakh MIDI file


carlthome:
I can open [this attached MIDI file](https://github.com/olemb/mido/files/526992/0b0812979e33c132046d24575ff810cf.mid.zip) in timidity and Ardour without issue, but mido is unable to parse it. What's special about it?

E.g.

``` python
from mido import MidiFile
MidiFile('0b0812979e33c132046d24575ff810cf.mid')
```

results in 

``` sh
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-51ef249dd91a> in <module>()
----> 1 MidiFile('0b0812979e33c132046d24575ff810cf.mid')

/home/carl/anaconda3/lib/python3.5/site-packages/mido/midifiles.py in __init__(self, filename, file, type, ticks_per_beat, charset)
    229         elif self.filename is not None:
    230             with io.open(filename, 'rb') as file:
--> 231                 self._load(file)
    232 
    233     def add_track(self, name=None):

/home/carl/anaconda3/lib/python3.5/site-packages/mido/midifiles.py in _load(self, file)
    266 
    267             for i in range(number_of_tracks):
--> 268                 self.tracks.append(self._read_track())
    269                 # Todo: used to ignore EOFError. I hope things still work.
    270 

/home/carl/anaconda3/lib/python3.5/site-packages/mido/midifiles.py in _read_track(self)
    351                 message = self._read_sysex()
    352             else:
--> 353                 message = self._read_message(status_byte)
    354 
    355             message.time = delta

/home/carl/anaconda3/lib/python3.5/site-packages/mido/midifiles.py in _read_message(self, status_byte)
    286         for byte in data_bytes:
    287             if byte > 127:
--> 288                 raise IOError('data byte must be in range 0..127')
    289         return build_message(spec, [status_byte] + data_bytes)
    290 

OSError: data byte must be in range 0..127
```

``` sh
pip show mido

---
Metadata-Version: 2.0
Name: mido
Version: 1.1.17
Summary: MIDI Objects for Python
Home-page: https://mido.readthedocs.io/
Author: Ole Martin Bjorndalen
Author-email: ombdalen@gmail.com
Installer: pip
License: MIT
Location: /home/carl/anaconda3/lib/python3.5/site-packages
Requires: 
Classifiers:
  Development Status :: 5 - Production/Stable
  Intended Audience :: Developers
  Natural Language :: English
[...]

PR #68, opened 2016-10-29
Added error mode option to MidiFile as discussed in issue #63.

olemb:
This passes an `error_mode` variable around in `read_track()`, `read_message()` etc.

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -20,6 +20,7 @@ import io
 import time
 import string
 import struct
+import warnings
 from ..messages import Message, SPEC_BY_STATUS
 from .meta import MetaMessage, build_meta_message, meta_charset
 from .meta import MetaSpec, add_meta_spec, encode_variable_int
@@ -30,6 +31,11 @@ from .tracks import MidiTrack, merge_tracks, fix_end_of_track
 DEFAULT_TEMPO = 500000
 DEFAULT_TICKS_PER_BEAT = 480
 
+
+class MidiFileError(IOError):
+    pass
+
+
 def print_byte(byte, pos=0):
     char = chr(byte)
     if char.isspace() or not char in string.printable:
@@ -38,7 +44,19 @@ def print_byte(byte, pos=0):
     print('  {:06x}: {:02x}  {}'.format(pos, byte, char))
 
 
-class DebugFileWrapper(object):
+class FileWrapper(object):
+    def __init__(self, file, error_mode='raise'):
+        self.file = file
+
+    def read(self, size):
+        return self.file.read(size)
+
+    def tell(self):
+        return self.file.tell()
+
+
+
+class DebugFileWrapper(FileWrapper):
     def __init__(self, file):
         self.file = file
 
@@ -68,6 +86,37 @@ def read_bytes(infile, size):
 
 def _dbg(text=''):
     print(text)
+
+
+def _handle_error(error_msg, error_mode):
+    if error_mode == 'raise':
+        raise MidiFileError(error_msg)
+    elif error_mode == 'warn':
+        warnings.warn(error_msg)
+    elif error_mode == 'debug':
+        _dbg(error_msg)
+    elif error_mode == 'ignore':
+        pass
+    else:
+        raise ValueError('unknown error mode {!r}'.format(error_mode))
+
+
+def _clip_data_bytes(data_bytes, error_mode):
+    """Clip data byte to 0..127.
+
+    Raises MidiFileError if error_mode =='raise'.
+    """
+
+    clipped_bytes = []
+    for byte in data_bytes:
+        if byte > 127:
+            _handle_error('out of range data byte in MIDI file', error_mode)
+            if error_mode != 'raise':
+                byte = 127
+
+        clipped_bytes.append(byte)
+
+    return clipped_bytes
 
 
 # We can't use the chunk module for two reasons:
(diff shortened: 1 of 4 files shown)


carlthome:
ETA on this being on PyPI?

Love mido, by the way! Wonderfully written package overall! :smiley:

olemb:
Thank you! I am happy to hear that you enjoy Mido. :)

Does the code look good? I kind of lost track of the discussion here so I'm not sure I implemented what you guys were asking for. :)

ETA, well it depends. This is written for 1.2 so unless we backport it to 1.1.x it will have to wait for the 1.2 release.

That said, 1.2 is very close to release ready. What remains:

* There's a bug in the RtMidi backend that causes deadlocks in some rare situations. I'm able to reproduce it so I just have to sit down and do some debugging.

* The documentation needs to be updated to reflect the API changes, such as the addition of `Message.from_bytes()` and friends.

I will set off some time every day to work on this. I appreciate any help I can get.

Thanks for the video. Over the last few years I've become more aware of the benefits of functional programming and keeping things stateless and side effect free, which is why I've started breaking `MidiFile` out into functions. Passing around `error_mode` is clearly the right thing to do but the way it came out it looked a bit tacked on and out of place. I was wondering if there's a better way to pass it around. Having it as the first argument would probably be even worse.

I guess it's not a big deal in any case. As this is not a part of the public API we just use something that works for now and change it later if we come up with something better. The important decision is whether or not to have an `error=` argument. The rest is internal.

olemb:
@carlthome, @craffel   Should I just go ahead and merge this?

(Hmm, the pull request got closed when I deleted the develop branch. Makes sense.)

mido/midifiles/midifiles.py
```diff
@@ -70,6 +88,37 @@ def _dbg(text=''):
     print(text)
 
 
+def _handle_error(error_msg, error_mode):
```
carlthome: I'm note sure error handling needs this encapsulation as [warnings](https://docs.python.org/2.7/library/warnings.html#the-warnings-filter) has builtin support for elevating warnings to exceptions and [logging](https://docs.python.org/2/library/logging.html#logging.Logger.debug) has support for different levels. 

Particularly as `_dbg` only prints to stdout at the moment (but you might have plans to expand on it in the future, I don't know). It would be more convenient to take in a Logger, probably.

mido/midifiles/midifiles.py
```diff
@@ -38,7 +44,19 @@ def print_byte(byte, pos=0):
     print('  {:06x}: {:02x}  {}'.format(pos, byte, char))
 
 
-class DebugFileWrapper(object):
+class FileWrapper(object):
+    def __init__(self, file, error_mode='raise'):
```
carlthome: Why does this take `error_mode`? It doesn't seem to use it.

mido/midifiles/midifiles.py
```diff
@@ -70,6 +88,37 @@ def _dbg(text=''):
...
+        warnings.warn(error_msg)
+    elif error_mode == 'debug':
+        _dbg(error_msg)
+    elif error_mode == 'ignore':
+        pass
+    else:
+        raise ValueError('unknown error mode {!r}'.format(error_mode))
+
+
+def _clip_data_bytes(data_bytes, error_mode):
```
carlthome: Nice! It's silly that this is the convention among MIDI software (Ardour, Logic, Timidity, etc.), but best be consistent.

mido/midifiles/midifiles.py
olemb: Ah, well spotted. Also, `FileWrapper` is never used. I don't know what I was thinking there. I'll remove it.

mido/midifiles/midifiles.py
olemb: This is where I get confused. :) Could you show me how you would implement it?

olemb: It would be a lot easier to understand in code.

carlthome: For example:

```python
import warnings

# Let's do our own warning type so it can be ignored without ignoring other modules warnings.
class MidiFileWarning(Warning):
    pass

# Use this whenever you want to display errors, instead of `_handle_error`.
warnings.warn('The error message', MidiFileWarning)
```

Then, either have a public API for setting the warning level:
```python
def set_error_level(error_mode):
    if error_mode == 'raise':
        warnings.filterwarnings('error', category=MidiFileWarning)
    elif error_mode == 'warn':
        warnings.filterwarnings('default', category=MidiFileWarning)
    elif error_mode == 'ignore':
        warnings.filterwarnings('ignore', category=MidiFileWarning) 
    else:
        raise ValueError('unknown error mode {!r}'.format(error_mode))
```
Or, preferably skip the if statements and just document that the user can call `warnings.filterwarnings(action, category=MidiFileWarning)` themselves!

Note that warnings introduces a global state per Python process so to be really correct, each function should use the context manager `warnings.catch_warnings` and take the `error_mode` argument, for example:

```python
def read_message(infile, status_byte, peek_data, delta, error_mode):
    # Stuff...

    # We want to print a warning with error_mode, so we temporarily change the warnings filters according to error_mode with a context manager.
    with warnings.catch_warnings():
        warnings.filterwarnings(error_mode, category=MidiFileWarning)
        warnings.warn('The error message', MidiFileWarning)

    # More stuff...
```

That might be a little overkill initially though, so perhaps just skip having `error_mode` entirely and leave it as a global state. Unless multiprocessing is expected, of course. :confounded:

PR #70, opened 2016-11-28
README.rst: fix usage of the "with" statement

ao2:
Fix the usage of the "with" python statement in one of the examples.

While at it also rename the "input" variable to "midi_input", to avoid having
a variable using the same name of a keyword of the language.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,8 +12,8 @@ designed to be as straight forward and Pythonic as possible:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


olemb:
I somehow didn't notice that when I updated the example. Thank you for fixing it.

I've merged your changes to my 1.1.x branch so they will be included if there is another 1.1 release. For 1.2 (in my develop branch) I've rewritten the example without the `with` clause to make it a little bit shorter.

PR #71, opened 2016-12-05
Add second2tick and tick2second utility functions

carlthome:
Just a proposal, and perhaps you don't think this is suited for mido @olemb, but I figured it would be neat to have utility functions alongside tempo2bpm and bpm2tempo for converting ticks to seconds and seconds to ticks, given a particular tempo and PPQN. Thoughts, @olemb? Also, did I count this right?

# TODO
 - [ ] Add unit tests

### DIFF
--- a/docs/midi_files.rst
+++ b/docs/midi_files.rst
@@ -186,56 +186,41 @@ Tempo and Beat Resolution
 
 .. image:: images/midi_time.svg
 
-Timing in MIDI files is all centered around beats. A beat is the same
-as a quarter note.
+Timing in MIDI files is centered around ticks and beats. A beat is the same as 
+a quarter note. Beats are divided into ticks, the smallest unit of time in 
+MIDI.
 
-Tempo is given in microseconds per beat, and beats are divided into
-ticks.
+Each message in a MIDI file has a delta time, which tells how many ticks have 
+passed since the last message. The length of a tick is defined in ticks per 
+beat. This value is stored as ``ticks_per_beat`` in MidiFile objects and 
+remains fixed throughout the song.
 
-The default tempo is 500000 microseconds per beat (quarter note),
-which is half a second per beat or 120 beats per minute. The meta
-message 'set_tempo' can be used to change tempo during a song.
 
-You can use :py:func:`bpm2tempo` and :py:func:`tempo2bpm` to convert
-to and from beats per minute. Note that :py:func:`tempo2bpm` may
-return a floating point number.
+MIDI Tempo vs. BPM
+^^^^^^^^^^^^^^^^^^
 
-Computations::
+Unlike music, tempo in MIDI is not given as beats per minute, but rather in 
+microseconds per beat.
 
-    beats_per_seconds = 1000000 / tempo
-    beats_per_minute = (1000000 / tempo) * 60
-    tempo = (60 / beats_per_minute) * 1000000
+The default tempo is 500000 microseconds per beat, which is 120 beats per 
+minute. The meta message 'set_tempo' can be used to change tempo during a song.
 
-Examples::
+You can use :py:func:`bpm2tempo` and :py:func:`tempo2bpm` to convert to and 
+from beats per minute. Note that :py:func:`tempo2bpm` may return a floating 
+point number.
 
-    2 == 1000000 / 500000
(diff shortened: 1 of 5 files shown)


olemb:
That sounds useful. I'll hopefully find some time to have a look at it tomorrow.

I think this should be available in the `mido` namespace or at least in `mido.midifiles`. `mido.midifiles.meta` is not supposed to be a part of the public API. (Perhaps it should be renamed to `_meta.py`?)

Why did I never think of using scientific notation in my tempo functions? `1e-6` is so much easier to read. :)

carlthome:
Please do! I'd love a proposal on how to unit test this properly by the way. Just compare to some known values?

olemb:
I'd like to switch to one of your functions in the `MidiFile` class. What I have today is:

    def _get_seconds_per_tick(self, tempo):
        return (tempo / 1000000.0) / self.ticks_per_beat

Feel free to make this change if you like.

For some reason I find tempo and time in MIDI files very confusing. I just can't keep the three units straight in my head so it took me a while to get the implementation right and I'll have to spend a little more time studying your functions to see what's going on.

Maybe you've seen it but I have a section on time and tempo in the docs:

    http://mido.readthedocs.io/en/latest/midi_files.html#tempo-and-beat-resolution

If you have ideas for improvements on this section I'd be very grateful. I feel like there's a lot of room for improvement.

Also, I didn't know you could do `from .module import (many, things, in, parenthesis)`. That's neat. I'll use that from now on.

carlthome:
Done! :smile: 

Uh, could you run the tests? Perhaps the repo could add [Travis](https://travis-ci.org/) so I don't mess anything up?

Agreed, MIDI files are very confusing. :confused: 

`(...)` is used throughout Python to allow line wrapping, and is not only for imports. Alternatively [one can append a backslash](https://github.com/carlthome/python-audio-effects/blob/master/tests/test_dsp.py#L8-L13) for continuing a line on the next one (like MATLAB's ellipsis).

carlthome:
Rewrote the documentation a little. I'd love some feedback.

olemb:
Your rewrite is a big improvement. Thank you. :)

All tests pass here.

I'll have to look into Travis. I'm currently using a git commit hook which runs ``pytest2 && python3`` so I can't commit until all tests pass.

Testing with a few key values should be enough. This is what I've done with ``tempo2bpm()`` and ``bpm2tempo()``.

This may be a bit nitpicky, but maybe the names should be plurals: ``def ticks2seconds(ticks, `` etc.? (On the other hand maybe they shouldn't. I'm not sure.)

I looked back at what I'm doing in ``MidiFile.__iter__()``. Every time the tempo changes I call ``get_seconds_per_tick(tempo, ticks_per_beat)`` which gives me a scale I can multiply with. Maybe this function could be exposed along with yours? (It may need a better name though.)

I see that your pull request targets my ``master`` branch which is where I keep the latest stable release. I will merge into ``develop`` which is currently heading towards 1.2.

Most pull requests I get target ``master``. I'm wondering if I've got this set up backwards. :)

olemb:
I'm getting some merge conflicts. I'll let you know when I've sorted them out and pushed to develop.

olemb:
OK, merge done.

olemb:
I renamed the first argument to `tick2second()` to `ticks`.

carlthome:
Ah, great! So as this has been merged by you to branch develop this pull request is actually completed then. :+1: 

Yes, it's common to let the latest dev branch be master, actually. There's no expectation on master to be stable so maybe set it up like that instead of having a separate develop branch. :koala:

carlthome:
Hmm, did we get some unit tests too?

olemb:
I'll probably switch development over to master then. There are currently two development branches though, 1.1.x and develop (1.2). There is no active development in 1.1.x beyond bug fixes so I guess 1.2 should be master?

olemb:
No unit tests yet. Would you like to write some? (I'm not sure I trust myself to do this given how many times I've messed up the time and tempo conversions.)

olemb:
I've taken your advice and moved development to the master branch. It is live now.

picturedots:
If you ended up here looking for the location of these functions, you can import them like

`from mido.midifiles import tick2second, second2tick, bpm2tempo, tempo2bpm`

PR #72, opened 2016-12-09
added missing init of base class for SocketPort

twobraids:
The class SocketPort neglected to call the super class ```__init__``` method.  This resulted in ```AttributeError: 'SocketPort' object has no attribute '_lock'``` errors.  Simply adding the call allows the class work correctly.


twobraids:
withdrawn - wrong destination branch - will resubmit at a later date

olemb:
Thanks for the fix. I missed that one.

No need to resubmit for such a small fix. I've merged it into my develop branch.

PR #73, opened 2016-12-14
Fix `with` clause in wrong order.

jmmcd:


olemb:
Thank you! I have already received a fix for this but I will credit you in the changelog.

PR #74, opened 2016-12-23
Fix the open_input sample code

bozakov:


olemb:
Thank you! I have already received two fixes for this but I will credit you in the changelog.

PR #78, opened 2017-01-21
Duck-Type style time checking for Numpy ints

calclavia:
Referring to https://github.com/craffel/pretty-midi/issues/114

Numpy ints cannot be properly checked with `isinstance(value, int)` in Python 3.

### DIFF
--- a/mido/messages.py
+++ b/mido/messages.py
@@ -133,14 +133,16 @@ def get_spec(type_or_status_byte):
 
 
 def check_time(time):
-    """Check type and value of time.
-    
-    Raises TypeError if value is not an integer or a float
-    """
-    if PY2 and isinstance(time, long):
-        return
-
-    if not (isinstance(time, int) or isinstance(time, float)):
+    valid = True
+    try:
+        int(time)
+    except:
+        valid = False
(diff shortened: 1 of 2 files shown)


olemb:
Looks good. I think the best solution is to use `numbers` as you have done. We could convert them to normal ints and floats but I think it's OK to keep them as Numpy numbers. They should behave the same as normal numbers.

One thing though: checking against `numbers.Number` will also accept complex numbers. That's probably not very useful. Could you change it to `numbers.Real`?

This also makes my test for `long` redundant leaving us with only that one test, which is a nice improvement.

Should we do the same for the other attributes?

calclavia:
@olemb I'm not very familiar with MIDO. I only changed the part that I use and have tested and I'm not sure about other attributes.

I'll change it to `numbers.Real`.

olemb:
OK, no problem. I'll do the other attributes.

olemb:
I've released 1.1.21 with support for Numpy for all attributes based on your implementation. I had trouble merging your pull request so I ended up making a new commit but I've credited you in the changelog.

Thank you for contributing this!

PR #81, opened 2017-02-16
Fix dll name for darwin platform

slegroux:
Hi,
I noticed there was a discrepancy in the dll_name variable for Darwin platform which prevented mido to load the portmidi dynamic library properly on OS X and made all my mido-based examples crash.


olemb:
Thanks!

I've released 1.1.24 with your fix.

PR #84, opened 2017-03-12
Fix links in docs

miketwo:


olemb:
Ah, cut and paste strikes again. Thanks!

Issue #99, opened 2017-06-26
Poll and non-blocking receive not returning (Virtual Port)


cjappl:
Hello,

Seeing an issue where the non-blocking `port.receive()` and `port.poll` never return. Both of which, if I'm reading the doc correctly, should be non-blocking and return None when no message is available. 

Here are some tests that reproduce the issue:
```
import mido
def test_mido_port_blocking_rec():
      with mido.open_ioport('foo', virtual=True) as p:
          p.receive(block=False)
  
def test_mido_port_blocking_poll():
      with mido.open_ioport('foo', virtual=True) as p:
          p.poll()
```
Neither or these tests ever return on my system.

Happy to provide more information for debugging, just let me know! 

Using python-rtmidi as backend, mido version 1.2.7, python 3.6.1

olemb:
This puzzled me for a while but I finally tracked it down to an error in the `IOPort` class.

`mido.open_input()` will open an input and an output port and wrap them in an `IOPort` object.

When you call `p.receive(block=False)` the `IOPort` object should forward it to `p.input`, but along the way it failed to pass on the `block` option causing the call to always block.

This was a pretty bad bug. Thanks for reporting!

cjappl:
Happy to help, when you upload the latest version to pypi I am happy to verify the fix and close the issue! Just let me know.

olemb:
Could you test with my master branch? I'd like to know that it works before I put it on PyPI.

cjappl:
Checked it, and it is good to go! The IOPort will successfully return None if it is non-blocking and there are no messages waiting.

Thanks for the quick fix.

olemb:
I've released 1.2.8.

While investigating this I found another case where ``IOPort.poll()`` would block so I've included a fix for that as well.

cjappl:
Great! Confirmed the package has the fix in it. Thanks again for the quick fix

PR #100, opened 2017-06-29
Added test for non-blocking receive, changed test format to class

cjappl:
Based on https://github.com/olemb/mido/issues/99, wanted to add a test for this behavior for the future.

I also cleaned up the testing infrastructure for the port class.


olemb:
Thanks! Nice to have some tests for this.

Issue #108, opened 2017-08-13
self.is_output should be self.is_input in portmidi backend


crackleware:
https://github.com/olemb/mido/blob/6d4152701e876050fbfc3eb14048314856d5be7f/mido/backends/portmidi.py#L108-L112

olemb:
Thanks!

I've merged the pull request from @JuantAldea which fixed the problem, but I've credited you as well in the changelog.

PR #109, opened 2017-08-15
Fix portmidi backend

JuantAldea:
Hello,

There's an issue on the port selection code of the PortMidi backend which is solved by the first commit of this PR.

The second commit changes the logic for selecting ports of _get_name_device. Give that each device has two properties (is_input, is_output), I believe the presented is much saner way of doing said selection.


JuantAldea:
Looking in the issue track, this fixes #108

olemb:
Thanks!

I agree. Your version of `_get_named_device()` is a lot clearer.

PR #110, opened 2017-08-16
Added commit to allow for midi values in files to exceed 127.

cjappl:
If "clip" is specified in midi file initializer, these values will be clipped to 127, otherwise, an error will be thrown. As referenced in issue #63

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -37,9 +37,10 @@ DEFAULT_TICKS_PER_BEAT = 480
 # Maximum message length to attempt to read.
 MAX_MESSAGE_LENGTH = 1000000
 
+
 def print_byte(byte, pos=0):
     char = chr(byte)
-    if char.isspace() or not char in string.printable:
+    if char.isspace() or char not in string.printable:
         char = '.'
 
     print('  {:06x}: {:02x}  {}'.format(pos, byte, char))
@@ -90,7 +91,7 @@ def _dbg(text=''):
 # 1. we may have mixed big and little endian chunk sizes. (RIFF is
 # little endian while MTrk is big endian.)
 #
-# 2. the chunk module assumes that chunks are padded to the neares
+# 2. the chunk module assumes that chunks are padded to the nearest
 # multiple of 2. This is not true of MIDI files.
 
 def read_chunk_header(infile):
@@ -98,7 +99,7 @@ def read_chunk_header(infile):
     if len(header) < 8:
         raise EOFError
 
-    # Todo: check for b'RIFF' and switch endian?
+    # TODO: check for b'RIFF' and switch endian?
 
     return struct.unpack('>4sL', header)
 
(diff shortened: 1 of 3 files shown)


olemb:
Looks good! I really appreciate that you've fixed little things you found along the way.

Some comments:

I see you've changed "Todo" to "TODO". Is that a more common way to write it? If we make the change we should probably update everywhere.

I forgot that you could do "not in". Neat and much more readable.

We could move the byte clipping into a `clip_data_bytes()` function. Not a big deal though.

At some point it would be nice to have more specific exceptions instead of `IOError` everywhere. `MidiParseError` or something would be more helpful (as a subclass of `IOError` or course). Same for ports (`PortNotFound` etc.). (It's on the roadmap somewhere.)

I think the "not all MIDI messages have running status" comment is out of date and can be removed.

The "`if (not msg.is_meta and raw[0] < 0xf0 and raw[0] == running_status_byte):`" is a bit long. I try to keep lines below 80 characters. Though if we keep it on one line the parenthesis can go. :)

I guess all that's missing now is the warning (and documentation).

Should I merge now or do you want to update the pull request?

mido/midifiles/midifiles.py
```diff
@@ -117,19 +118,28 @@ def read_file_header(infile):
...
 
-    # Subtrac 1 for status byte.
+    # Subtract 1 for status byte.
     size = spec['length'] - 1 - len(peek_data)
     data_bytes = peek_data + read_bytes(infile, size)
 
-    for byte in data_bytes:
-        if byte > 127:
-            raise IOError('data byte must be in range 0..127')
+    adjusted_bytes = []
```
craffel: You can replace line 131-138 with
```Python
if clip:
    data_bytes = [byte for byte in data_bytes if byte < 127 else 127]
```

mido/midifiles/midifiles.py
```diff
@@ -334,8 +345,8 @@ def _load(self, infile):
                 if self.debug:
                     _dbg('Track {}:'.format(i))
 
-                self.tracks.append(read_track(infile, debug=self.debug))
-                # Todo: used to ignore EOFError. I hope things still work.
+                self.tracks.append(read_track(infile, debug=self.debug, clip=self.clip))
```
craffel: @olemb do you enforce an 80-character width limit (PEP-8)?

mido/midifiles/midifiles.py
cjappl: Nice! That's much cleaner, thank you.

mido/midifiles/midifiles.py
cjappl: Fixed it! That's the one thing I have my linter turned off on, thanks for catching it :)

cjappl:
@olemb I'll update this one right now!

> Looks good! I really appreciate that you've fixed little things you found along the way.

No problem! Maybe after this one I'll do a quick "clean up" pull request. I don't know if you know the tool [flake8](http://flake8.pycqa.org/en/latest/index.html#quickstart) but it's my preferred linter. It'll yell at you if anything is outside PEP8 and makes it really easy to clean up.

> I see you've changed "Todo" to "TODO". Is that a more common way to write it? If we make the change we should probably update everywhere.

Let me do this as a part of the "cleanup" review, just so I don't clutter this one. I know in my development environment, it boldly highlights TODO, but not Todo. Just makes them stand out better.

olemb:
Thanks!

Let's go with TODO. I've had a look around and that seems to be more or less standard. We could also consider changing some of them to FIXME where that applies.

I'll have a look at flake8.

I should use linters more. I ran pyflakes it helped me find lots of little things like unused imports: 4d96942dcedba46d8ab4c35d03fcef215a2a0ed2

mido/midifiles/midifiles.py
olemb: @craffel Yes, I enforce an 80-character limit.

PR #112, opened 2017-08-18
Cleaned codebase using flake8. Also switched instances of Todo -> TODO

cjappl:
No functional changes were made, just layout/PEP8 formatting

### DIFF
--- a/mido/__init__.py
+++ b/mido/__init__.py
@@ -43,12 +43,12 @@ Parsing MIDI streams:
     parse_all(bytes) -- parse all messages bytes
     Parser -- MIDI parser class
 
-Parsing objects serialized with str(message):                                  
+Parsing objects serialized with str(message):
 
     parse_string(string) -- parse a string containing a message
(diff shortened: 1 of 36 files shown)


olemb:
Thanks!

olemb:
One of the changes broke Python 2 support. I've fixed it in a57a99451c0be0e38bbb3b520fa2e59a6160d6a9.

olemb:
I just remembered you can pass a tuple of types to `isinstance()`: d47ad03b49d4cc456c01efe436e7042c9abab4c5

(I also didn't like having the two functions inside an if/else so I rewrite it like this. I think it's a bit cleaner.)

Otherwise great commit. :)

cjappl:
Awesome! Yeah highly recommend checking into using "flake8", you could even potentially include it in the pre-commit step of "Contributing" if you have such a document.

Also, do you know how I get added to "Contributors" on this project? Is that automatic by github? I'd love to say I've "officially" contributed! :)

Issue #102, opened 2017-07-10
possible unit conversion bug


superbock:
I am not completely sure, since I am by no means an expert for MIDI, but I think your conversion between bpm and tempo (i.e. `bpm2tempo` and `tempo2bpm`) does not take the time signature into account.

If I am correct, MIDI usually refers to the tempo in ticks per quarter note. Thus depending on the time signature, the duration of a beat may vary. The functions work only correctly for 4/4 time signature.

__Edit:__ updated the issue with a description, the initial issue was empty. Sorry.
__Edit 2:__ I think the same problem exists in all unit conversions in `midifiles.units`.

olemb:
Hi Sebastian,

Apologies for the late reply. I've been in vacation.

MIDI tempo is defined by microseconds per beat (quarter note). Time signature should not factor in as a beat has the same length in any time signature. For example a 2/4 measure would last half as long as the 4/4 measure. (2 beats vs. 4 beats.) Odd signatures like 2/3 are a bit trickier to compute but should also be based on the length of a quarter note.

I may be wrong though. Do you have an example of a time signature where this is not correct?

superbock:
Hi Ole, 

Yes, MIDI tempo is in microseconds per quarter note and for the default time signature of 4/4 everything is just fine. You are also right that a 2/4 bar lasts half as long as a 4/4 measure since each bar contains a different number of beats (2 vs. 4) and each beat has the same duration (a quarter note)

The problem, however, is that this is only true for 4/4 time signature. If the denominator changes (btw. 2/3 is not possible in MIDI, since the denominator can only be a power of 2), this also changes the length of a beat. If it is a x/2 time signature each beat has the length of a half note, i.e. two quarter notes. Thus the conversion from beats (in bpm) to MIDI tempo (in ticks per quarter note) has to consider the denominator of the time signature. At least this is my (sowmhow limited) knowledge about MIDI tempo, but it matches [this answers on music.stackexchange.com](https://music.stackexchange.com/a/24145).

I added updated versions of unit conversions [to my code](https://github.com/CPJKU/madmom/pull/308/files#diff-c50455c4f06264e1275f413634df9d71R19) so you can have a look at them. The updated unit conversion functions also required a fix to `mido.MidiFile.__iter__` in order to consider the time signature events (similar to the handling of tempo events). My `MIDIFile` class inherits from `mido.MidiFile` and contains this fix as well. I am happy to provide a PR with those fixes.

[...]

PR #114, opened 2017-08-20
Fix BPM <-> MIDI tempo conversions

superbock:
This PR addresses #102 and contains the fixes needed to correctly consider the time signature when converting between BPM (beats per minute) and MIDI tempo (microseconds per quarter note).

Before being able to be merged, these points should be discussed/addressed:

- [ ] `ticks_per_beat` is a bit misleading since it rather should be `ticks_per_quarter_note`. Renaming the variable, however, may break existing code, so this is a rather bad option. Maybe adding a new variable (and deprecate the old one, but keep it around for backwards compatibility) is a better approach.
- [ ] Tests for Python 2 fail at the moment since Python 2 and 3 round differently. This was revealed by adding tests for corner cases.
- [ ] Extend the documentation to reflect all introduced changes, especially update the image in `midi_files.rst`.
- [ ]  Add more descriptive docstrings, but which docstring format should be used?

### DIFF
--- a/mido/midifiles/units.py
+++ b/mido/midifiles/units.py
@@ -20,27 +20,23 @@ def second2tick(second, ticks_per_beat, tempo):
     return second / scale
 
 
-def bpm2tempo(bpm):
-    """Convert beats per minute to MIDI file tempo.
+def bpm2tempo(bpm, time_signature=(4, 4)):
+    """Convert BPM (beats per minute) to MIDI file tempo (microseconds per
+    quarter note).
 
-    Returns microseconds per beat as an integer::
-
-        240 => 250000
-        120 => 500000
-        60 => 1000000
+    Depending on the chosen time signature a bar contains a different number of
+    beats. These beats are multiples/fractions of a quarter note, thus the
+    returned BPM depend on the time signature.
     """
-    # One minute is 60 million microseconds.
-    return int(round((60 * 1000000) / bpm))
+    return int(round(60 * 1e6 / bpm * time_signature[1] / 4.))
 
 
-def tempo2bpm(tempo):
-    """Convert MIDI file tempo to BPM.
+def tempo2bpm(tempo, time_signature=(4, 4)):
+    """Convert MIDI file tempo (microseconds per quarter note) to BPM (beats
+    per minute).
 
-    Returns BPM as an integer or float::
-
-        250000 => 240
-        500000 => 120
-        1000000 => 60
+    Depending on the chosen time signature a bar contains a different number of
+    beats. These beats are multiples/fractions of a quarter note, thus the
+    returned tempo depends on the time signature.
     """
-    # One minute is 60 million microseconds.
-    return (60 * 1000000) / tempo
+    return 60 * 1e6 / tempo * time_signature[1] / 4.
--- a/mido/midifiles/units.py
+++ b/mido/midifiles/units.py
@@ -1,9 +1,9 @@
 def tick2second(tick, ticks_per_beat, tempo):
     """Convert absolute time in ticks to seconds.
 
-    Returns absolute time in seconds for a chosen MIDI file time
-    resolution (ticks per beat, also called PPQN or pulses per quarter
-    note) and tempo (microseconds per beat).
+    Returns absolute time in seconds for a chosen MIDI file time resolution
+    (ticks/pulses per quarter note, also called PPQN) and tempo (microseconds
+    per quarter note).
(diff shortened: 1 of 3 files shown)


mido/midifiles/units.py
```diff
@@ -17,7 +17,7 @@ def second2tick(second, ticks_per_beat, tempo):
     note) and tempo (microseconds per beat).
     """
     scale = tempo * 1e-6 / ticks_per_beat
-    return second / scale
+    return int(round(second / scale))
```
carlthome: This change will break a program of mine where I let the user decrease the PPQ of MIDI files.

```py
ticks = int(round(mido.second2tick(message.time, ppq, tempo)))
assert ticks > 0 or message.time == 0, "New PPQ is too low."
```

superbock: It would be good to see which part of the assert fails (I am not sure if this makes any difference, though), then it's easier to discuss this further.

The rationale behind this change is that "the MIDI standard" requires QQN to be an integer. And until now this function could return a float value.

I am not sure why ticks should be greater 0 or message timings should not be 0. Both are legitimate values as far as I can see. What's the intention behind these checks?

The problem seems to be that different Python versions handle the /-operator differently. Anyways, I see that this change breaks code (although I never thought it would), but IMHO the question remains if your assertion assumes the "right thing".

carlthome: > The rationale behind this change is that "the MIDI standard" requires QQN to be an integer. And until now this function could return a float value.

How should we handle the case when second will be truncated to zero for the given PPQ and tempo (or heavily shifted in time, `ceil` doesn't really help)?

> What's the intention behind these checks?

The assertion checks that there will be at least one tick with the given `ppq` when `message.time` is larger than 0.0 seconds. 

When PPQ is chosen too low there could be messages that don't move the tick count at all even though they should, which causes a slight global drift AFAIK.

The problem stems from MAPS having ridiculously high PPQ (https://github.com/craffel/pretty-midi/issues/112), while just setting it to 960 is fine because no message actually seems to require that much time resolution to fit into a piano roll (although the messages might shift just slightly forward and/or backwards in time, which I don't bother to check).

carlthome: Btw, what does QQN stand for?

superbock: Typo, TQN = ticks per quarter note.

superbock: High PPQ (e.g. in the MAPS dataset) should be fine, since they result in large tick values. On the other hand, having low PPQ values _is_ a (known) problem, but I am still not sure how this is affected by the proposed change.

> The assertion checks that there will be at least one tick with the given ppq when message.time is larger than 0.0 seconds.

Ok, but I don't see how moving the `int(round(...))` thing inside the function should alter the result since you apply exactly the same rounding and casting to int by yourself. Are you sure that this change is the culprit and not something else?

carlthome: > Ok, but I don't see how moving the int(round(...)) thing inside the function should alter the result since you apply exactly the same rounding and casting to int by yourself.

Haha, you're absolutely right... Sorry. :laughing: 

image

Anyway, could we have something like this internally maybe?

```python
if second != 0 and tick == 0:
    warnings.warn(f"{second} seconds is zero ticks at {ppq} PPQ.")
```

carlthome:
Status?

mido/midifiles/units.py
```diff
@@ -12,35 +12,31 @@ def tick2second(tick, ticks_per_beat, tempo):
...
+def tempo2bpm(tempo, time_signature=(4, 4)):
+    """Convert MIDI file tempo (microseconds per quarter note) to BPM (beats
+    per minute).
 
-        250000 => 240
-        500000 => 120
-        1000000 => 60
+    Depending on the chosen time signature a bar contains a different number of
+    beats. These beats are multiples/fractions of a quarter note, thus the
+    returned tempo depends on the time signature.
```
mgrachten: I think it is good to mention here that what constitutes a beat is defined by the denominator of the time signature, and give an example: in 2/4, 3/4, 4/4, ... the beat is a quarter note. In 3/8, 6/8, ... the beat is an eight note, etc.

rdoursenaud (changes requested):

rdoursenaud:
I have added the fix I was suggesting in #454 and merged your changes.

Thank you for your contribution!

superbock:
Thanks a lot!

PR #115, opened 2017-08-20
Add different time units to MidiFile

superbock:
**Note:** This PR adds the enhancements mentioned in #102 to `MidiFile`. It is based on PR #114, i.e. #114 needs to be merged first.

# Content
Add different time units and relative/absolute timimg capability to `MidiFile`.

`MidiFile.unit` can hold a string indicating which unit should be used to display event timings. Possible values are `ticks`, `seconds`, `beats` and various shortcuts thereof (`t`, `s`, `sec`, and `b`).

Furthermore the timings can be viewed as absolute values or relative to the preceeding event as indicated by `MidiFile.timing`. This attribute can also hold a string, possible values are `absolute` (shortcuts: `a`, `abs`) or `relative` (shortcuts: `r`, `rel`).

To support timing information in beats, time signature event handling was added to `MidiFile.__iter__` and the needed unit conversion functions to `midifiles.units`.

### Minor modifications

These modifications are split into separate commits which can be skipped if wanted:

- moved constants to `midifiles.units`,
- remove the obsolete `get_seconds_per_tick` function,
- add default arguments to unit conversion functions.

Before being able to be merged, these points should be discussed/addressed:

- [ ] Extend the documentation to reflect the changes in `MidiFile`.
- [ ] Import the new `tick2beat` and `beat2tick` into the main namespace?
- [ ] Include `tick2beat` and `beat2tick` in the API docs.
- [ ]  Add docstrings, but which docstring format to use?
- [ ] Renaming `ticks_per_beat` to `ticks_per_quarter_note` in the unit conversion functions?


superbock:
Comments welcome, thus ping @olemb.

cjappl:
I cast my vote for [numpydoc](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt). It hasn't failed me yet!

cjappl:
Did you run the tox tests? 

You can do so by getting the latest revision then:

`pip install -e .[dev]` 
`tox`

The first one is only necessary to make sure you have all the dev packages installed before you run tox.

cjappl:
also +1 from me for adding a bunch of unit tests, that's great!!

superbock:
Of course I ran the tests, but as said in #114 (and copied into this PR's TODO items) some tests fails due to different rounding in Python 2/3. My question was on what functionality is wanted. Either the rounding could be unified or the tests could be edited to check for the current Python version and test accordingly.

Either way it would be extremely helpful for all if there were some sort of continuous integration (e.g. TravisCI).

But I also wanted to get a more general feedback on this PR. Is this functionality wanted in mido? If yes, what changes to the PR are needed. If no I add them to the classes inherited from mido in my own projects. Then, still #114 should be merged, since it contains bug fixes.

carlthome:
> Either way it would be extremely helpful for all if there were some sort of continuous integration (e.g. TravisCI).

https://github.com/olemb/mido/pull/129

superbock:
Interesting. Can you explain how you checked this in order to repeat (and fix) it. Furthermore it would be interesting which commit adds this slowing down. Is #114 affected as well?

superbock:
I guess I figured it out: I introduced a `timing` parameter which must be 'relative' instead of 'absolute' since MIDI messages have relative timing. Should be fixed now. Thanks for pointing this out!

However, this reminds me on how difficult it is to change some code without proper tests. I can write a test for this but it requires a MIDI file to be tested. Are there any suggestions on which properties such a file should have or does something like a "standard" MIDI test file exist?

Jimmusic:
Many thanks for that I will give it a try.  Original problem is difficult to explain but when playing a song that begins with a slow tempo it takes ages for the first note to play.  In the case of the attached file about 11 secs but on a daw it starts instantly.  I wondered if it was something to do with temp and time sig so tried your changes.  I also think that it might be a delta time calc of some sort that is calculating the time to next message incorrectly when the first note is a rest.
[ffcommon.zip](https://github.com/olemb/mido/files/1569674/ffcommon.zip)

Jimmusic:
The latest changes fixed the slow motion - Great work.  There still remains the problem of a long time to get to the first note when playing the example file as opposed to the short wait when playing in a DAW.
This problem is in the original mido so not related to your enhancement.  I will report it there.
This is not related to the __iter__  and merge_tracks. it is after that.
By the way if it is any use: The above __iter__ and for msg in merge_tracks(self.tracks) does not work well when looping a drum loop so I changed the class to  __init__  to include
self.mergeset = []
and the last line of def _load to have
self.mergeset = merge_tracks(self.tracks)
and the __iter__ to have
for msg in self.mergeset
Loops now work perfectly

I will print out the time delta for each message to try and find out why it takes so long to get to the first note of the above file and similar slow tempo midi files starting with rests.

Again many thanks
PS what do your changes let me do ? Can I find out what beat I am on ?  That would be great.

Regards
Jim

superbock:
The long delay at the beginning is unrelated to the changes in this PR. Caching the merged tracks can indeed solve the problem -- or at least lower its extend. I'll look into it and incorporate this suggestion as well, possibly in a separate PR.

The `unit` parameter enables you to change the time units to be either in seconds (default), ticks (MIDI internal timing) or beats. So the `time` attribute of the individual messages are measured in the selected unit. Furthermore, with the `timing` parameter you can choose if you want absolut timings or relative to  the preceding message.

Jimmusic:
Caching the merged tracks only solves the problem of creating loops such as drum loops where they have to be seamless. The problem I am experiencing with long time to first note when a midi file starts that has quite a lot of control changes and a slow tempo  and or rests at beginning of song is that the first note on message seems to have a very large time Delta. In the file I uploaded the first one has 12 secs. Most of proceeding channel messages have 0 time Delta but some get 3secs. Saw shows them all as 0. The odd deltas seem to follow where but I only had a brief look. Will try to describe better later. This is not isolated to this midi.

Jimmusic:
I missed out seems to follow sysex messages

carlthome:
Status?

superbock:
I am not sure what the status of this PR is. I am still waiting for any kind of response to the questions I raised.

docs/midi_files.rst
```diff
@@ -181,46 +181,55 @@ The ``time`` attribute is used in several different ways:
...
 
 Converting Between Ticks and Seconds
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-To convert from MIDI time to absolute time in seconds, the number of beats per 
-minute (BPM) and ticks per beat (often called pulses per quarter note or PPQ, 
-for short) have to be decided upon.
+To convert from MIDI time to absolute time in seconds, the tempo (either
+in number of beats per minute (BPM) or microseconds per quarter note, see
+above) and ticks per per quarter note have to be decided upon.
```
carlthome: `see above` should be a reST hypertext reference to the relevant section instead.

mido/midifiles/midifiles.py
```diff
@@ -277,28 +274,19 @@ def write_track(outfile, track):
...
-    #    (500000 / 1000000) / 480 == 0.5 / 480 == 0.0010417
-    #
-    return (tempo / 1000000.0) / ticks_per_beat
-
-
 class MidiFile(object):
     def __init__(self, filename=None, file=None,
                  type=1, ticks_per_beat=DEFAULT_TICKS_PER_BEAT,
-                 charset='latin1',
+                 unit='seconds', timing='relative', charset='latin1',
```
carlthome: Will this break backwards compatibility or were these defaults implicit before (I think yes, just wanna double check)?

mido/midifiles/midifiles.py
```diff
@@ -373,18 +361,40 @@ def __iter__(self):
...
-                delta = tick2second(msg.time, self.ticks_per_beat, tempo)
+                if self.unit.lower() in ('t', 'ticks'):
+                    delta = msg.time
+                elif self.unit.lower() in ('s', 'sec', 'seconds'):
+                    delta = tick2second(msg.time, self.ticks_per_beat, tempo)
+                elif self.unit.lower() in ('b', 'beats'):
+                    delta = tick2beat(msg.time, self.ticks_per_beat,
+                                      time_signature)
+                else:
+                    raise ValueError("`unit` must be either 'ticks', 't', "
```
carlthome: The shorthands should be removed ('t', 's', 'b' are not obvious) as they provide very little benefit at a big cost. `MidiFile(unit='beats')` is a lot cleaner and easier to search in the docs after.

I.e. 

```
if self.unit.lower() == 'ticks':
```
is better.

mido/midifiles/midifiles.py
```diff
@@ -373,18 +361,40 @@ def __iter__(self):
...
+                                     "'seconds', 's', 'beats', 'b', not %s." %
+                                     self.unit)
             else:
                 delta = 0
+            # Convert relative time to absolute values if needed
+            if self.timing.lower() in ('a', 'abs', 'absolute'):
```
carlthome: Same here. Shorthands are harder to search for help with for users etc.

mido/midifiles/units.py
```diff
@@ -1,46 +1,73 @@
...
     """Convert absolute time in ticks to seconds.
 
-    Returns absolute time in seconds for a chosen MIDI file time
-    resolution (ticks per beat, also called PPQN or pulses per quarter
-    note) and tempo (microseconds per beat).
+    Returns absolute time in seconds for a chosen MIDI file time resolution
+    (ticks/pulses per quarter note, also called PPQN) and tempo (microseconds
+    per quarter note).
     """
+    # Note: both tempo (microseconds) and ticks are per quarter note
```
carlthome: Remove comment.

mido/midifiles/units.py
```diff
@@ -1,46 +1,73 @@
...
     """Convert absolute time in seconds to ticks.
 
-    Returns absolute time in ticks for a chosen MIDI file time
-    resolution (ticks per beat, also called PPQN or pulses per quarter
-    note) and tempo (microseconds per beat).
+    Returns absolute time in ticks for a chosen MIDI file time resolution
+    (ticks/pulses per quarter note, also called PPQN) and tempo (microseconds
+    per quarter note).
     """
+    # Note: both tempo (microseconds) and ticks are per quarter note
```
carlthome: Remove comment.

mido/midifiles/units.py
```diff
@@ -1,46 +1,73 @@
...
+    #       thus the time signature is irrelevant
     scale = tempo * 1e-6 / ticks_per_beat
     return tick * scale
 
 
-def second2tick(second, ticks_per_beat, tempo):
+def second2tick(second, ticks_per_beat=DEFAULT_TICKS_PER_BEAT,
```
carlthome: "Convert absolute time in seconds to ticks." > "Round absolute time in seconds to ticks." IMO, or something of that sort to clarify that the conversion is lossy.

mido/midifiles/midifiles.py
superbock: Yes, these were implicit before, so this change shouldn't break anything. There was a bug in a previous version which changed the default behaviour, but this is fixed ([see comment above](https://github.com/olemb/mido/pull/115#issuecomment-352399362)). However, I can't guarantee this because there are no tests covering this aspect.

andresbrocco:
Any advancements?! I can't wait for this improvement

superbock:
Unfortunately, I haven't heard back from the developers after I made some changes to the PR. I will update the code and rebase on the current master branch.

mido/midifiles/midifiles.py
rdoursenaud: IMHO the supported units should be defined by an `Enum` with sensible names.
This would make the code clearer and simplify types verification.

mido/midifiles/midifiles.py
rdoursenaud: Again, an `Enum` would solve this.
No need to check for multiple strings variations.

Issue #105, opened 2017-07-25
KeyError: (8, 0) when opening a Midi file, probably an error in the file that is not handled?!


schtibe:
I try to open a MIDI file that I received from a customer, and it ends with following error:

```
----> 1 mido.MidiFile("/tmp/PianoFeelThisMoment.mid")

~/.pyenv/versions/3.5.3/envs/energyday/lib/python3.5/site-packages/mido/midifiles/midifiles.py in __init__(self, filename, file, type, ticks_per_beat, charset, debug)
    300         elif self.filename is not None:
    301             with io.open(filename, 'rb') as file:
--> 302                 self._load(file)
    303 
    304     def add_track(self, name=None):

~/.pyenv/versions/3.5.3/envs/energyday/lib/python3.5/site-packages/mido/midifiles/midifiles.py in _load(self, infile)
    335                     _dbg('Track {}:'.format(i))
    336 
--> 337                 self.tracks.append(read_track(infile, debug=self.debug))
    338                 # Todo: used to ignore EOFError. I hope things still work.
    339 

~/.pyenv/versions/3.5.3/envs/energyday/lib/python3.5/site-packages/mido/midifiles/midifiles.py in read_track(infile, debug)
    209 
    210         if status_byte == 0xff:
--> 211             msg = read_meta_message(infile, delta)
    212         elif status_byte in [0xf0, 0xf7]:
    213             # Todo: I'm not quite clear on the difference between

~/.pyenv/versions/3.5.3/envs/energyday/lib/python3.5/site-packages/mido/midifiles/midifiles.py in read_meta_message(infile, delta)
    163     length = read_variable_int(infile)
    164     data = read_bytes(infile, length)
--> 165     return build_meta_message(type, data, delta)
    166 
    167 

~/.pyenv/versions/3.5.3/envs/energyday/lib/python3.5/site-packages/mido/midifiles/meta.py in build_meta_message(type, data, delta)
    440 
    441         # This adds attributes to msg:
--> 442         spec.decode(msg, data)
    443 
    444         return msg

~/.pyenv/versions/3.5.3/envs/energyday/lib/python3.5/site-packages/mido/midifiles/meta.py in decode(self, message, data)
    383         key = signed('byte', data[0])
    384         mode = data[1]
--> 385         message.key = _key_signature_decode[(key, mode)]
    386 
    387     def encode(self, message):

KeyError: (8, 0)

```

[...]

PR #117, opened 2017-08-21
Added KeySignatureError and tests. Also added bidict dependency.

cjappl:


mido/midifiles/meta.py
```diff
@@ -11,6 +11,7 @@
      - expose _key_signature_encode/decode?
 """
 from __future__ import print_function, division
+from bidict import bidict
```
cjappl: I can definitely remove this if need be. I felt at least a little weird adding a dependancy for it.

This is a great package, especially for this "encode/decode" thing you have going on here with the dictionaries.

The bidict basically can have keys as values, and values as keys, so you can access the "inverse" of the dictionary with "my_bidict.inv", similar to what was being done in "reverse_table".

Again, happy to revert out this bidict change, just wanted to show what it was capable of!

cjappl:
From #105

cjappl:
@olemb Sorry to bug you, but I, unfortunately, wrote this on my master branch. I wanted to clear this up before moving on to the tox/dev dependencies task.

Cheers!

olemb:
Sure thing.

I ran into a problem though. `./setup.py` won't run unless `bidict` is installed:

```
Traceback (most recent call last):
  File "./setup.py", line 4, in <module>
    import mido
  File "/home/olemb/src/mido/mido/__init__.py", line 100, in <module>
    from .midifiles import (MidiFile, MidiTrack, merge_tracks,
  File "/home/olemb/src/mido/mido/midifiles/__init__.py", line 1, in <module>
    from .meta import MetaMessage, UnknownMetaMessage, KeySignatureError
  File "/home/olemb/src/mido/mido/midifiles/meta.py", line 14, in <module>
    from bidict import bidict
ImportError: No module named bidict
```
This is because it imports things like `__version__` and `__author__` from the top level module, which also imports `meta.py`.

This means pip will fail as well so you can't actually install Mido unless bidict is already installed. There must be a fix for this that doesn't involve updating version etc. in two places.

cjappl:
Got a solution, will submit it with the other PR

olemb:
Cool!

cjappl:
see #123

PR #119, opened 2017-08-29
Add defaults to second2tick and tick2second

carlthome:
This is just a suggestion (and I'm not 100% sure this is a good idea) but it's pretty convenient to have Mido's default PPQ and tempo already set for unit conversion to make it easier to use absolute time when interacting with Mido.

### DIFF
--- a/mido/midifiles/units.py
+++ b/mido/midifiles/units.py
@@ -1,4 +1,7 @@
-def tick2second(tick, ticks_per_beat, tempo):
+from .midifiles import DEFAULT_TEMPO, DEFAULT_TICKS_PER_BEAT
+
+
+def tick2second(tick, ticks_per_beat=DEFAULT_TICKS_PER_BEAT, tempo=DEFAULT_TEMPO):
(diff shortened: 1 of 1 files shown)


cjappl:
Seems convenient, I guess one of the problems comes with if you're working with a bunch of files with a bunch of different tempos. I don't know if this is a cleaner approach but we could do a different function like "default_ticks_to_secs" or something that could use the default params?

Not sure if I'm in love with that either though...

superbock:
I think this is a good idea, I proposed the same in #115 (see bc2988ca) 😏 

Regarding @cjappl's concern, I don't see a problem why adding default values should cause any problems. Existing code doesn't work without setting these params, new code can rely on the default values.

@carlthome: maybe have a look at #115, this PR adds a mechanism to set both the desired time unit and whether you want to retrieve the timings as absolute or relative values.

carlthome:
Sweet. Let's close this one then.

By the way, congratulations on the thesis! I'm just getting started reading it. :heart_eyes:

PR #120, opened 2017-09-03
 Re-added import ctypes.byref to fix portmidi backend

jkevin:
This was fixed in commit d051be32e4f6ee52f4570bcfed445aae20a92526 but then broken again with codebase cleaning in df6d05a6abcf6139ca31715dd3ed5450b2d98e96


olemb:
Thanks!

flake8 reported this as an unnecessary import, but it didn't consider that `portmidi.py` would need it. Something to keep in mind in the future.

PR #123, opened 2017-09-19
Added tox tests. Added dev dependencies. Updated docs. 

cjappl:
...fixed issue with bidict dependency being imported before setup.py could be run.

From now on (as noted in contributing.rst) we should require people to run the `tox` command before submission. This will check:

- The full install in a new env will run successfully
- all tests pass
- manifest is checked
- docs will be able to be built
- no PEP8 violations

In the future, we could consider adding code coverage as well. 

There is no way to have tox install python versions for you, unfortunately. You have to install them yourself before the command is run. I recommend we require at least 27 and 35 pass, and any of the others (32, 36) are just nice-to-haves. That way we aren't saddling contributors with 4 python versions.


cjappl:
Also, I was wondering how I could be added as a contributor to the project? I wasn't sure if there was something I should be doing for my commits to "count" towards that! 

I'm new to the contributing to open source world

olemb:
Looking good!

I'm going to have to look into Tox a bit more and play around with it. The output is a bit overwhelming. I wonder if there's a way to show only errors and have it stop at the first error.

Nice solution to the version number. I ran into a problem with the docs though. If you change the version number in setup.py (to 1.2.10000 for example) and run `./setup.py docs` it will still say `1.2.8` in the docs. If you run `tox` and then `./setup.py docs` again you get `1.2.10000`. I don't know what's going on there.

Another problem: I put `tox` to my pre-commit hook but now I can't commit because I don't have Python 3.2 and 3.6. How is this working for you?

You're now in third place on the contributors page. :) Good work!

cjappl:
> Another problem: I put tox to my pre-commit hook but now I can't commit because I don't have Python 3.2 and 3.6. How is this working for you?

There are two possible solutions for this. First, we could use the "skip_missing_interpreters" command, which will skip interpreters instead of failing. 

Second, we could just enforce py27, py35 and py36, and just remove py32. 

I'd personally want the second one, as I think using the skip_missing command will just end up in people not running the important stuff.

more info here:
http://tox.readthedocs.io/en/latest/config.html#confval-skip_missing_interpreters=BOOL

> I'm going to have to look into Tox a bit more and play around with it. The output is a bit overwhelming. I wonder if there's a way to show only errors and have it stop at the first error.

I have a solution for the verbosity, nothing fancy, just passing in -q to the commands that accept such a thing. I agree it's way better than it was when I submitted, the output isn't streaming by recklessly! 

> Nice solution to the version number. I ran into a problem with the docs though. If you change the version number in setup.py (to 1.2.10000 for example) and run ./setup.py docs it will still say 1.2.8 in the docs. If you run tox and then ./setup.py docs again you get 1.2.10000. I don't know what's going on there.

Seems like I may have found a fix for this as well. It seems as if pkg_control just gets the installed version number, as opposed to the updated version number. If you install the package every time you build the docs, it'll be fixed.

> You're now in third place on the contributor's page. :)

I'm finally contributing to a project :'). Brings a tear to my eye!


I'll submit another PR for these little things after we discuss the versioning ideals. Do you think we should use the skip_missing_interpreters, or just make sure that people are testing with minimally 27, 35 and 36?

olemb:
Great to have you on board! :)

I'm not sure. For now I just put `tox -e py27,py35` in my pre-commit hook.

I also tried adding `-q`s. Works really well:

```
[tox]
envlist = py27, py32, py35, py36

[testenv]
commands = pip install -q -e .[dev] 
           py.test -q . -rs 
           check-manifest
           sphinx-build -q docs {envtmpdir} -E
           flake8 mido --ignore=F401
```

 It's OK if it's not perfect right away. We'll keep tweaking it.

PR #124, opened 2017-09-21
Added options to quiet down tox output. Fixed issue where pkg versionwasn't being updated correctly

cjappl:
As discussed in #123


olemb:
Very nice!

I removed the extra pip install step. It's not needed anymore now that I put __version__ back into the package.

PR #129, opened 2017-10-02
Add CI

carlthome:
@olemb, it would be nice to run the tests on pull requests. Could you go to www.travis-ci.org and activate this repo with their CI (it's free for public repos)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: python
+python:
+- "2.7"
+- "3.6"
+cache: pip
+install: pip install -e .[dev]
+script: pytest


carlthome:
Also [tox is sweet](https://github.com/olemb/mido/blob/master/tox.ini), but I don't know if it will work with concurrent CI builds. It's nice to test each python version in parallel, which is why I've made the .travis.yml like this. @cjappl maybe has comments?

carlthome:
And is py32 and py35 still necessary targets or is py36 enough?

superbock:
I definitely vote against Python 3.2 since it is EOL and doesn't even get source-only security updates (see https://www.python.org/dev/peps/pep-0392/#lifespan). Same applies to 3.3 (https://www.python.org/dev/peps/pep-0398/#lifespan).

hugovk:
You can certainly can use tox with Travis CI, just google it to find plenty of guides. But this simple way is a good start.

Generally I'd say it's a good idea to test on all Python versions you support, for example, 2.7, 3.4, 3.5 and 3.6.

Whilst on the topic, it looks some versions numbers could be updated here (and would be great to do it in this PR):
* https://github.com/olemb/mido#requirements
* https://github.com/olemb/mido/blob/master/tox.ini#L2
* https://github.com/olemb/mido/blob/master/setup.py#L70-L71

olemb:
Thanks!

Looks it's working:

     https://travis-ci.org/olemb/mido/

cjappl:
https://pypi.python.org/pypi/tox-travis

Seems like it's pretty easy to get tox up and running in this instead of pytest, can we do something like the example here?

carlthome:
Just curious, what would be the pros of using tox instead of pytest on Travis (that already uses different Python versions in the build matrix)?

hugovk:
A benefit is that as a byproduct, it is also testing the tox.ini file. Another is that the same thing is being run on Travis as locally (assuming you use tox locally).

cjappl:
https://github.com/olemb/mido/blob/master/tox.ini

If you check out our tox.ini it's doing a bunch of stuff on top of pytest. It has a step for checking that the style is enforced, as well as the MANIFEST is still right.

It's basically an "all in one" test for the package, it's nice and easy to run and put any checks you'd want in there!

PR #131, opened 2017-10-09
Added routine to get total message count from MidiFiles. 

scottmudge:
I thought it would be nice to be able to get the total message count of a midi file from MidiFile, without having to load the data back into a Parser and parse it all over again.

Useful for creating arrays of static size to manipulate midi data (e.g., with numpy).

Apologies for the gratuitous commits... never use notepad's tabs.

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -302,7 +302,7 @@ class MidiFile(object):
         self.charset = charset
         self.debug = debug
         self.clip = clip
-        self.tempo = DEFAULT_TEMPO
+
         self.tracks = []
 
         if type not in range(3):
@@ -378,18 +378,19 @@ class MidiFile(object):
         if self.type == 2:
             raise TypeError("can't merge tracks in type 2 (asynchronous) file")
 
+        tempo = DEFAULT_TEMPO
         for msg in merge_tracks(self.tracks):
             # Convert message time from absolute time
             # in ticks to relative time in seconds.
             if msg.time > 0:
-                delta = tick2second(msg.time, self.ticks_per_beat, self.tempo)
+                delta = tick2second(msg.time, self.ticks_per_beat, tempo)
             else:
                 delta = 0
 
             yield msg.copy(time=delta)
 
             if msg.type == 'set_tempo':
-                self.tempo = msg.tempo
+                tempo = msg.tempo
 
     def play(self, meta_messages=False):
         """Play back all tracks.
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -302,7 +302,7 @@ class MidiFile(object):
         self.charset = charset
         self.debug = debug
         self.clip = clip
-
+        self.tempo = DEFAULT_TEMPO
         self.tracks = []
 
         if type not in range(3):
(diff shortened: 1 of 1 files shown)


mido/midifiles/midifiles.py
```diff
@@ -365,6 +368,13 @@ def length(self):
                              ' for type 2 (asynchronous) file')
 
         return sum(msg.time for msg in self)
+		
+    def msg_count(self):
```
cjappl: I think this may be better defined as overriding the __len__ of the file? https://docs.python.org/2/reference/datamodel.html#object.__len__

This could just be as simple as:
len(my_file)

Maybe more pythonic. If we wanted msg_count, it could just be an alias for len

olemb: I agree. ``__len__()`` is the way to go here.

It's best to compute this on demand since the message count can change as you add and remove tracks and messages. I propose something like this (off the top of my head):

```
    def __len__(self):
        return sum(len(track) for track in self.tracks)
```

scottmudge: I tried that but it returned the length as a metric of time, not messages. I suppose I could just iterate through the messages wherever I import MidiFile, but it seemed weird not to have it as an accessible property.

scottmudge: Ah, that's what I was looking for. I didn't want to loop through each time, but I'm relatively new to the python ecosystem still (C++ guy).

cjappl: welcome to python, any questions, never hesitate to reach out!

scottmudge: Yeah the __len__ implementation worked well. Strangely the __len__ implementation results in 5 more total messages than the iterative mess I put here for this midi file:
https://github.com/scottmudge/orchestrion/raw/master/data/midis/tf_in_d_minor.mid

But there are also 5 named tracks, so perhaps it's a difference between length (N+1) and total accessible elements (N).

scottmudge:
Alright just used the \_\_len\_\_ suggestion. Feel free to roll it in if you like. I find it useful, but I can just implement it on my end too.

Thanks for the help!

mido/midifiles/midifiles.py
olemb: I understand your confusion about the `length` attribute. I should probably have chosen a more specific name for it such such as `total_time`.

cjappl:
This seems like something nice to have around. I guess my only concern would be
"what do people think is the length of a midi file?"

This could mean a number of things:
number of notes in the file (as implemented)
length in time of the file
length in bytes of the file

Just playing devils advocate here, not sure what the right move is.

scottmudge:
Yeah I suppose it's a toss up between being more pythonic at the cost of ambiguity, or less ambiguous at the cost of potentially more code. I'd be fine with an explicit method for number of messages too. 

What's the consensus?

olemb:
@cjappl  Yes, that ambiguity is what has kept me from adding `__len__()` in the past.

I tried a quick implementation and ran into another source of confusion:

```
>>> import mido
>>> mid = mido.MidiFile('Weightless.mid')
>>> len(mid)
16709
>>> len(list(mid))
16690
```

So... `len()` returns more messages than you will get when you iterate over the file. This seems strange, but the reason is that `__iter__()` doesn't just iterate over the messages, it merges them into one track (removing all the `end_of_track` messages) and converts time stamps to seconds.

I would like to deprecate `__iter__()` into `mid.play(sleep=False)`. I'm not sure how much code that would break.

cjappl:
Hmm, yeah at the very least len(list(blah)) and len(blah) should return the same thing

olemb:
Definitiely, but if `len(mid)` returns the same as `len(list(mid)` it will still be confusing why this is different from the actual number of messages in the file.

I don't see how this can be solved without either removing or redefining `__iter__().

mido/midifiles/midifiles.py
olemb: Those 5 messages are `end_of_track` messages that are filtered out when `self.__iter__()` merges all tracks into one for playback.

cjappl:
Due to that, it may be better to go back to the original name "msg_count" for this one, and punt on the iter/len thing?

scottmudge:
Sounds good, I'll revert back to the old nomenclature to avoid ambiguity. It'll be a couple hours until I get back to my machine.

cjappl:
To be clear, I defer to @olemb for the final decision, I'd wait to hear what he says. I'm just a hanger-on contributor :)

scottmudge:
Doing some other work on the repo for the time being. Will loop back.

PR #135, opened 2017-10-26
Pass `block` parameter through in `MultiPort._receive()`

rec:
I saw this while going over the source!

I haven't actually encountered a problem due to what I perceive is an issue on this line - it might be the correct behavior for some reason I haven't figured out.

If it's right, let me know and I'll change the commit to be a comment explaining it.

Thanks for an excellent library!

### DIFF
--- a/mido/ports.py
+++ b/mido/ports.py
@@ -369,7 +369,7 @@ class MultiPort(BaseIOPort):
     def _receive(self, block=True):
         self._messages.extend(multi_receive(self.ports,
                                             yield_ports=self.yield_ports,
-                                            block=False))
+                                            block=block))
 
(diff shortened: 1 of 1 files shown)


cjappl:
This seems good to me! not sure why it would be a parameter if it's going to be ignored.

rec:
Ping!

cjappl:
@olemb has been MIA for a little bit, he may be on vacation or something

olemb:
Sorry for taking so long to get back to you!

That was indeed a mistake. Thanks for the fix!

rec:
No problem at all, thanks for doing this!  :-)

PR #138, opened 2017-11-06
Fixed spelling errors in docs

cjappl:
Nothing too extreme here, just realized one error so fixed a bunch


olemb:
Some interesting misspellings there. Thanks!

PR #149, opened 2018-02-01
Added from_bytes() method to MetaMessage class

gulaki:
I've tested this only on a Windows 10 system on python 3.6.1. Tested it on all possible kinds of
MetaMessage messages. I don't know if there's some standards that I should have abided by in
writing the function. I think this is quite useful and I was missing this feature.


rdoursenaud:
I have added tests for your new method and will merge your changes soon.

Thank you for your contribution!

Issue #160, opened 2018-05-01
[question] does playback of MidiFile drift?


belm0:
Since time.sleep won't be exact, won't it cause the MIDI playback to drift over time?  This would seem especially likely if the messages are fine-grained.
```
for msg in MidiFile('song.mid'):
    time.sleep(msg.time)
    if not msg.is_meta:
        port.send(msg)
```
It seems like the time given to sleep should be based on the wall time since playback started and the logical tick position within the MIDI file.

belm0:
E.g. see this info on sleep accuracy.  It's especially poor on Windows.
https://stackoverflow.com/questions/1133857/how-accurate-is-pythons-time-sleep

If say, the MIDI data has some fine-grained expression like a pitch bend event every 5-10 milliseconds, the position error in Mido's play loop would quickly accumulate.

It should be straightforward to replace the current `play()` implementation with something based on wall time and tick position.  I'll probably try a pull request.

belm0:
I made a modification to the example scripts to demonstrate drift.

https://github.com/belm0/mido/commit/ea258f794b492db17f1fc8d1b1365ccc0efdbb40

The generated midi file now includes fine-grained expression, and the granularity can be controlled by command line.  We can observe drift increase with data granularity, and hence the number of sleep calls.   Results on OS X (which has a very accurate sleep implementation):
```
$ ./create_midi_file.py 20 && ./play_midi_file.py test.mid
...
play time: 2.72 s (expected 2.50)

$ ./create_midi_file.py 5 && ./play_midi_file.py test.mid
...
play time: 2.84 s (expected 2.50)
```

belm0:
With play() fix:
```
./create_midi_file.py 5 && ./play_midi_file.py test.mid
...
play time: 2.50 s (expected 2.50)
```

PR #161, opened 2018-05-02
Improve MidiFile.play to avoid time drift.

belm0:
Addresses #160

Also modify midifile examples to include fine-grained events
and report time drift.

### DIFF
--- a/examples/midifiles/play_midi_file.py
+++ b/examples/midifiles/play_midi_file.py
@@ -8,6 +8,7 @@ Run with (for example):
 """
 import sys
 import mido
+import time
 from mido import MidiFile
 
 filename = sys.argv[1]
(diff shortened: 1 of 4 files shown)


examples/midifiles/play_midi_file.py
```diff
@@ -18,9 +19,13 @@
...
 with mido.open_output(portname) as output:
     try:
-        for message in MidiFile(filename).play():
+        midifile = MidiFile(filename)
+        t0 = time.time()
+        for message in midifile.play():
             print(message)
             output.send(message)
+        print 'play time: %.2f s (expected %.2f)' % (
+            time.time() - t0, midifile.length)
```
bast: Here I would suggest to use a print function compatible with Python 3, possibly also using `.format()`.

bast: Thank you!

PR #164, opened 2018-06-11
eval-able repr() for Message, MidiTrack, MidiFile

### LINKED ISSUE #162: proper __repr__ for MidiFile, etc.
Per Python docs, `__repr__()` output "should normally look like a valid Python expression that can be used to recreate an object with the same value".

Currently for MidiFile, repr() gives the same result as str(), which is just a summary including number of tracks, etc.  Likewise MidiTrack gives a summary with number of messages.

One place a proper repr() would be immediately useful is for tests, e.g. to confirm that code changes don't alter generated MIDI messages, tracks, or files.  It would also facilitate viewing differences between MIDI data in text form.

With a proper repr(), it would also be trivial to implement MidiFile `__eq__` (see #150).

belm0:
Fixes #162.

Example:
```Python console
>>> from mido import Message, MidiTrack, MidiFile
>>> m = Message('note_on', channel=1, note=2, time=3)
>>> m
Message('note_on', note=2, velocity=64, channel=1, time=3)
>>> t = MidiTrack([Message('note_on', channel=1, note=2, time=3),
... Message('note_off', channel=1, note=2, time=3)])
>>> t
MidiTrack([
  Message('note_on', note=2, velocity=64, channel=1, time=3),
  Message('note_off', note=2, velocity=64, channel=1, time=3)])
>>> f = MidiFile(type=1, ticks_per_beat=123, tracks=[t])
>>> f
MidiFile(type=1, ticks_per_beat=123, tracks=[
  MidiTrack([
    Message('note_on', note=2, velocity=64, channel=1, time=3),
    Message('note_off', note=2, velocity=64, channel=1, time=3)])
])
>>> f2 = eval(repr(f))
>>> f2
MidiFile(type=1, ticks_per_beat=123, tracks=[
  MidiTrack([
    Message('note_on', note=2, velocity=64, channel=1, time=3),
    Message('note_off', note=2, velocity=64, channel=1, time=3)])
])
```

Design decisions:
  * output of str() for each of these classes is left as it was
  * repr is canonicalized to the actual MIDI data as much as possible.  Hence MidiFile parameters like "filename", "debug" etc. are ignored, and Frozen variants are represented in mutable form.

### DIFF
--- a/mido/frozen.py
+++ b/mido/frozen.py
@@ -3,9 +3,14 @@ from .midifiles import MetaMessage, UnknownMetaMessage
 
 
 class Frozen(object):
-    def __repr__(self):
-        text = super(Frozen, self).__repr__()
+    def __str__(self):
+        text = super(Frozen, self).__str__()
         return '<frozen {}'.format(text[1:])
+
+    def __repr__(self):
+        # canonicalize to mutable objects
+        r = super(Frozen, self).__repr__()
+        return r[len('Frozen'):]
 
     def __setattr__(self, *_):
         raise ValueError('frozen message is immutable')
--- a/mido/messages/messages.py
+++ b/mido/messages/messages.py
@@ -54,6 +54,15 @@ class BaseMessage(object):
         """
         return cl(**data)
 
+    def __repr__(self):
+        d = self.dict()
+        msg_type = d.pop('type')
+        items = getattr(d, 'iteritems', d.items)
+        return "%s('%s', %s)" % (
+            type(self).__name__,
+            msg_type,
+            ', '.join('%s=%s' % item for item in items()))
+
     @property
     def is_realtime(self):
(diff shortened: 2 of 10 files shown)


olemb:
I will look over the code a bit and make some small changes but otherwise it looks good. I agree that this is a better repr() format.

We'll also have to update all examples in the documentation.

Thanks for the help!

olemb:
I made some final changes to the code (mostly switching from % syntax for `.format()`) and updated the docs to your new repr() format.

I also removed the __str__() methods from MidiFile and MidiTrack. There's really no use for them. repr() has all we need here.

I'd like to thank you for taking the time to implement this! The new format is a great improvement. Especially seeing the content of tracks and files is much better than the old summary. Having used this new repr() for a while the angle bracket format just looks wrong to me now.

PR #173, opened 2018-10-03
Mention mailing list and maintenance status

bast:
We look for developers who would like to join or take over maintenance of this code.


olemb:
Looks good. Thanks!

PR #178, opened 2018-10-31
add comment on the time drift difference and trim code

bast:
@belm0 @olemb please have a look.

I trimmed it a bit and added a comment. Hopefully for the better but do not hesitate to ask for changes.

Initially I thought about taking this optional argument passing away:
https://github.com/mido/mido/blob/master/examples/midifiles/create_midi_file.py#L22
... but I left it.

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -401,14 +401,14 @@ class MidiFile(object):
         safely modify them without ruining the tracks.
         """
         t0 = time.time()
-        combined_time = 0.0
+        input_time = 0.0
 
         for msg in self:
-            combined_time += msg.time
+            input_time += msg.time
 
             # we compute the difference between the nominal combined time
             # and the actual time spent at this point
-            duration_to_next_event = combined_time - (time.time() - t0)
+            duration_to_next_event = input_time - (time.time() - t0)
 
             if (duration_to_next_event > 0.0):
                 time.sleep(duration_to_next_event)
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -399,18 +399,19 @@ class MidiFile(object):
 
         You will receive copies of the original messages, so you can
         safely modify them without ruining the tracks.
-
         """
-        sleep = time.sleep
-        now = time.time
-        t0 = now()
-        file_time = 0
+        t0 = time.time()
+        combined_time = 0.0
 
         for msg in self:
-            file_time += msg.time
-            duration_to_next_event = file_time - (now() - t0)
-            if (duration_to_next_event > 0):
-                sleep(duration_to_next_event)
+            combined_time += msg.time
+
+            # we compute the difference between the nominal combined time
+            # and the actual time spent at this point
+            duration_to_next_event = combined_time - (time.time() - t0)
+
+            if (duration_to_next_event > 0.0):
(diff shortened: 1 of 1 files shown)


mido/midifiles/midifiles.py
```diff
@@ -399,18 +399,19 @@ def play(self, meta_messages=False):
...
+        combined_time = 0.0
 
         for msg in self:
-            file_time += msg.time
-            duration_to_next_event = file_time - (now() - t0)
-            if (duration_to_next_event > 0):
-                sleep(duration_to_next_event)
+            combined_time += msg.time
+
+            # we compute the difference between the nominal combined time
```
belm0: this comment doesn't seem of high value since it's reiterating the expression

I think the meaning of the expression is apparent, especially with the rename to `input_time` I suggested.  However, it could be clarified slightly with an intermediate variable for delta time (again avoiding a comment in favor of inherently descriptive code).
```
playback_time = time.time() - t0
duration_to_next_event = input_time - playback_time
```

mido/midifiles/midifiles.py
```diff
@@ -399,18 +399,19 @@ def play(self, meta_messages=False):
...
+            # and the actual time spent at this point
+            duration_to_next_event = combined_time - (time.time() - t0)
+
+            if (duration_to_next_event > 0.0):
```
belm0: I understand use of `0.0` for the variable initialization, since it's a cheap form of type annotation.  However in this expression no variable is being annotated, and it's perfectly valid to compare a float to an int, so `0` is the nominal form.

mido/midifiles/midifiles.py
olemb: @belm0, I was struggling with this code until I saw your version here. This makes perfect sense to me. There's a `playback_time` that's trying to keep up with the `input_time` via `sleep()`.

`playback_time` is a very good name. Maybe the other variable could be `target_time`? I think `input_time` is fine though.

mido/midifiles/midifiles.py
olemb: I have no strong opinion on this, but I think `0.0` helps a little bit extra here since it makes it clear that we're dealing with seconds and not MIDI ticks (which would always be integers).

olemb: @bast, can you remove the parentheses from around the expression?

    if duration_to_next_event > 0.0:

bast: Will do - I was programming something else in Fortran at the same time and it polluted me :-)

mido/midifiles/midifiles.py
```diff
@@ -399,18 +399,18 @@ def play(self, meta_messages=False):
...
+        input_time = 0.0
 
         for msg in self:
-            file_time += msg.time
-            duration_to_next_event = file_time - (now() - t0)
-            if (duration_to_next_event > 0):
-                sleep(duration_to_next_event)
+            input_time += msg.time
+
+            playback_time = time.time() - t0
```
olemb: `t0` looks a bit out of place. Would `start_time` be a better name? (Maybe there's some convention here I'm unaware of that speaks against it.)

bast: No problem for me to rename `t0` to `start_time`. Upcoming ...

olemb: Cool! Thanks.

PR #182, opened 2018-12-13
Fixed broken flake8 invocation by moving its options to a config file.

olemb:
Also fixed a couple of line that flake8 complained about.

Tox runs again without any errors (at least locally). I've silenced a few errors. We may want to go back and look into them in the future.


olemb:
Thanks for merging! It's good to get the tests working again.

PR #191, opened 2019-02-23
Only copy ports list once in ports.multi_receive

rec:
This is tiny but I kept seeing it as I read through the code and eventually... :-D

In `ports.multi_receive`, you _do_ want to make a copy of the incoming `ports` list, but only once, not each time through the loop!  So I moved it out of the loop.

### DIFF
--- a/mido/ports.py
+++ b/mido/ports.py
@@ -384,9 +384,9 @@ def multi_receive(ports, yield_ports=False, block=True):
 
     If block=False only pending messages will be yielded.
     """
+    ports = list(ports)
     while True:
         # Make a shuffled copy of the port list.
-        ports = list(ports)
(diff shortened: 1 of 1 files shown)


olemb (approved):

olemb:
That makes more sense. I don't know why I put that inside the loop.

Thanks for the fix!

rec:
Not a problem at all!

Sorry I didn't reply to your previous email, got distracted.  I'm actually working more on numpy generated digital audio these recent months, but I'm sure I'll get back to MIDI at some point.

If you want to discuss future plans for Mido, either to get someone to code review or just to bounce ideas of someone, hit me up - tom@swirly.com

Thanks again for this excellent and aesthetically pleasing library!

PR #199, opened 2019-06-28
Correct project URLs for python-rtmidi and rtmidi_python

SpotlightKid:


SpotlightKid:
Ping...

BTW: the CI failure is not due to this change. There is a flake8 error here:

    mido/messages/messages.py:15:9: F901 'raise NotImplemented' should be 'raise NotImplementedError'

SpotlightKid:
Ping.

olemb:
I've merged your changes and hope to have an update out within the next few weeks. Unfortunately the online docs will have to be wrong for a little bit longer.

I don't know how the tests missed the `NotImplemented` before. I guess they've added something to `flake8`. Thanks for the help though.

PR #252, opened 2020-11-25
Updated changelog.

olemb:


olemb:
This is not a good way to update the change log. Hopefully there's a better way than creating a pull request every time.

Issue #256, opened 2020-12-17
Resource leak and crash in get_output_names()


TheOtherDays:
This code:
```python
import mido

for i in range(0, 100):
	print(i)
	p = mido.get_output_names()
	del p
```
crashes on linux after 32 iterations (v 1.2.9, tested on multiple computers)

Output:
```python
0
1
2
3
[...]
30
31
32
ALSA lib seq_hw.c:466:(snd_seq_hw_open) open /dev/snd/seq failed: Cannot allocate memory
Traceback (most recent call last):
  File "src/_rtmidi.pyx", line 818, in rtmidi._rtmidi.MidiIn.__cinit__
RuntimeError: MidiInAlsa::initialize: error creating ALSA sequencer client object.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "p.py", line 5, in <module>
    p = mido.get_output_names()
  File "/home/eric/.local/lib/python3.8/site-packages/mido/backends/backend.py", line 175, in get_output_names
    devices = self._get_devices(**self._add_api(kwargs))
  File "/home/eric/.local/lib/python3.8/site-packages/mido/backends/backend.py", line 163, in _get_devices
    return self.module.get_devices(**self._add_api(kwargs))
  File "/home/eric/.local/lib/python3.8/site-packages/mido/backends/rtmidi.py", line 52, in get_devices
    input_names = set(rtmidi.MidiIn(rtapi=rtapi).get_ports())
  File "src/_rtmidi.pyx", line 820, in rtmidi._rtmidi.MidiIn.__cinit__
rtmidi._rtmidi.SystemError: MidiInAlsa::initialize: error creating ALSA sequencer client object.
```

RtMidi's ``get_ports()`` do leak some resources and must be explicitely freed (see https://github.com/SpotlightKid/python-rtmidi/issues/60#issuecomment-609437630).

The following patch prevents the leak and the crash:

```diff
--- mido/backends/rtmidi.py	2020-12-17 22:54:07.115227374 +0100
+++ /tmp/rtmidi.py	2020-12-17 22:53:48.254984192 +0100
@@ -48,9 +48,10 @@
     devices = []
 
     rtapi = _get_api_id(api)
-
-    input_names = set(rtmidi.MidiIn(rtapi=rtapi).get_ports())
-    output_names = set(rtmidi.MidiOut(rtapi=rtapi).get_ports())
+    mi = rtmidi.MidiIn(rtapi=rtapi)
[...]

PR #264, opened 2020-12-18
Fix RtMidi leaks when using MidiIn or MidiOut (#256)

TheOtherDays:
Some remarks:

- https://github.com/mido/mido/blob/master/mido/backends/rtmidi.py#L119 and https://github.com/mido/mido/blob/master/mido/backends/rtmidi.py#L184
still use MidiIn and MidiOut

- examples might need to be fixed, I'm not sure

- other rtMidi-based backends might need some work too

### DIFF
--- a/mido/backends/rtmidi.py
+++ b/mido/backends/rtmidi.py
@@ -49,8 +49,10 @@ def get_devices(api=None, **kwargs):
 
     rtapi = _get_api_id(api)
 
-    input_names = set(rtmidi.MidiIn(rtapi=rtapi).get_ports())
-    output_names = set(rtmidi.MidiOut(rtapi=rtapi).get_ports())
+    mi = rtmidi.MidiIn(rtapi=rtapi)
+    mo = rtmidi.MidiOut(rtapi=rtapi)
+    input_names = set(mi.get_ports())
+    output_names = set(mo.get_ports())
 
     for name in sorted(input_names | output_names):
         devices.append({'name': name,
@@ -58,6 +60,8 @@ def get_devices(api=None, **kwargs):
                         'is_output': name in output_names,
                         })
 
+    mi.delete()
+    mo.delete()
     return devices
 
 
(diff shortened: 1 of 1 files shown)


olemb (approved):
Looks good.

olemb:
Thanks!

* the `MidiIn` and `MidiOut` that are created here are cleaned up in `PortCommon._close()`. That's why I added that last `delete()`. (Sorry about the multiple inheritance. It makes the code a bit hard to follow.)
* as long as the code `port.close()` this will take care calling `delete()`. If the port is not closed the `MidiIn` or `MidiOut` object will hang around anyway.
* that's a good point. We should check if the `rtmidi_python` backend has the same problem. I'll add a new issue.

TheNetStriker:
I've also just tested this fix and it works. I always had this problem because I watch for usb events and then attach the midi ports. Every time after 32 reconnects the following error occured:

ALSA lib seq_hw.c:466:(snd_seq_hw_open) open /dev/snd/seq failed: Cannot allocate memory
Traceback (most recent call last):
  File "src/_rtmidi.pyx", line 818, in rtmidi._rtmidi.MidiIn.__cinit__
RuntimeError: MidiInAlsa::initialize: error creating ALSA sequencer client object.

Here is my test script if anyone is interested:

```
import mido
import time

def searchMidiDevice():
    port_names = mido.get_input_names()
    for name in port_names:
        if "X-Touch One" in name:
            return name
    return None

midiPortname = searchMidiDevice()

inputMidi = None
outputMidi = None

try:
    i = 1
    while i < 100:
        if inputMidi != None:
            inputMidi.close()
            inputMidi = None
        if outputMidi != None:
            outputMidi.close()
            outputMidi = None

        print(i)
        i += 1
        inputMidi = mido.open_input(midiPortname)
        outputMidi = mido.open_output(midiPortname)
        time.sleep(2)
except OSError as e:
    print("connectMidi() Error: {}".format(e))
```

cyberic99:
@olemb do you plan to release a new version anytime soon?

this leak/crash can really be a blocker for some uses

thanks!

Issue #266, opened 2020-12-20
Move automated testing from Travis to GitHub Actions


bast:
Working on a PR ...

bast:
My plan:
- add GH Actions (but do not remove the rest)
- observe it for few days
- only later remove .travis.yml, tox.ini, and the Travis badge

bast:
Example: https://github.com/mido/mido/actions/runs/434433524

olemb:
Nice!

Kind of look forward to some tests failing now so I can see what that looks like. :-D

PR #280, opened 2020-12-20
test using GitHub Actions

bast:
this is part of #266


bast:
Note that this does not do anything until it arrives on `master` but every later pull request will then be tested. I have tested this on my fork.

olemb (approved):
OK, let's go.

PR #292, opened 2021-03-04
Allow non virtual ports with client_name

georgeharker:
It would be super handy if non-virtual ports could be created with a client name - RtMidi supports this.

(i have some stuff that doesn't work with virtual ports, and still needs a nice naming scheme)

How about the current behavior as default, with the option to explicitly not create a virtual port if virtual is specified too?


rdoursenaud:
I doubt this supported on all 3 platforms (Linux, Microsoft Windows and Apple Mac OS).
Would you be willing to check?

PR #320, opened 2021-07-03
Type hinting

phuang1024:
Added type hinting to function parameter and return types.

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -313,13 +313,13 @@ class MidiFile(object):
 
         if type not in range(3):
             raise ValueError(
-                'invalid format {} (must be 0, 1 or 2)'.format(format))
+                'invalid format {} (must be 0, 1 or 2)'.format(type))
 
         if tracks is not None:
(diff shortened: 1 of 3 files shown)


michaelkaeppler:
What's the state of this? I'm working on a project that utilizes mido and static type checking. I would love to see type hinting implemented in mido.

mido/midifiles/midifiles.py
```diff
@@ -294,13 +295,9 @@ def get_seconds_per_tick(tempo, ticks_per_beat):
...
 
 class MidiFile(object):
-    def __init__(self, filename=None, file=None,
-                 type=1, ticks_per_beat=DEFAULT_TICKS_PER_BEAT,
-                 charset='latin1',
-                 debug=False,
-                 clip=False,
-                 tracks=None
-                 ):
+    def __init__(self, filename: str = None, file: IO[bytes] = None, type: int = 1,
```
michaelkaeppler, line 298: I thought that it has to be `filename: Optional[str] = None` and the like.
Does the current state pass a `mypy` run?

Issue #329, opened 2021-08-06
callback with arguments


folkertvanheusden:
Hi,

Please add argument-option to the callback so that the same callback function can be used for multiple sessions.

With this, a callback function can be connected to multiple ports without the need for global variables and difficult locking stuff.

thank you

rdoursenaud:
Per the reply in #329, I consider the request invalid.
Feel free to reopen if I'm mistaken.

PR #331, opened 2021-08-06
Implement issue #329 (callback with arguments for input-ports)

folkertvanheusden:
Introduces 'callback_w_args' callback-parameter for input-ports.
This implements issue #329

### DIFF
--- a/examples/callback_w_args/callback_w_args.py
+++ b/examples/callback_w_args/callback_w_args.py
@@ -1,0 +1,15 @@
+import mido
+import queue
+import time
+
+rtmidi = mido.Backend('mido.backends.rtmidi')
+
+def myfunc(msg, q):
+    print(f'Got message: {msg}')
+    q.put(msg)
+
+q = queue.Queue()
+input_ = rtmidi.open_input('testing', virtual=True, callback_w_args=(myfunc, q))
+
+while True:
+    time.sleep(1)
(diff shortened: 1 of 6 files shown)


examples/callback_w_args/callback_w_args.py
```diff
@@ -0,0 +1,15 @@
...
+import time
+
+rtmidi = mido.Backend('mido.backends.rtmidi')
+
+def myfunc(msg, q):
+    print(f'Got message: {msg}')
+    q.put(msg)
+
+q = queue.Queue()
+input_ = rtmidi.open_input('testing', virtual=True, callback_w_args=(myfunc, q))
```
smiszym, line 12: There's no need to implement anything in mido to achieve this. The standard way would be to use [`functools.partial`](https://docs.python.org/3.9/library/functools.html#functools.partial) to create a new callable that has some parameters set to something. (if I understand correctly what you want to achieve). Like this:

```py
input_ = rtmidi.open_input('testing', virtual=True, callback=partial(myfunc, q=q))
```

folkertvanheusden: Thank you! I did not know about "partial".

folkertvanheusden:
Change is not required.

examples/callback_w_args/callback_w_args.py
smiszym: Good question. Indeed you could use partial with Thread too. I guess args is there just for convenience as starting a Thread with some arguments is quite common.

PR #341, opened 2021-10-10
Support running status

m-vo:
We're now storing the last status byte whenever a message with at least one data byte is composed. If a message starts with a data byte this status byte will get injected. This is according to the MIDI spec: [running status](http://midi.teragonaudio.com/tech/midispec/run.htm).

So the following two inputs are considered equal now:

    0x9a, 3, 127, 0x9a, 3, 64, 0x9a, 3, 0
    0x9a, 3, 127,       3, 64,       3, 0   

I'm not sure about the right base branch, I'll rebase accordingly if needed.

### DIFF
--- a/mido/tokenizer.py
+++ b/mido/tokenizer.py
@@ -12,6 +12,7 @@ class Tokenizer(object):
         """Create a new decoder."""
 
         self._status = 0
+        self._last_status = 0
         self._bytes = []
         self._messages = deque()
         self._datalen = 0
@@ -59,7 +60,12 @@ class Tokenizer(object):
             if len(self._bytes) == self._len:
                 # Complete message.
                 self._messages.append(self._bytes)
+                self._last_status = self._status
                 self._status = 0
+        elif self._last_status:
+            # Running status
+            self._feed_status_byte(self._last_status)
+            self._feed_data_byte(byte)
         else:
             # Ignore stray data byte.
             pass
--- a/tests/test_tokenizer.py
+++ b/tests/test_tokenizer.py
@@ -52,4 +52,12 @@ def test_sysex_inside_sysex():
 
 def test_stray_data_bytes():
     """Data bytes outside messages should be ignored."""
-    assert tokenize([0, 1, 0x90, 2, 3, 4, 5, 0xf8, 6]) == [[0x90, 2, 3], [0xf8]]
+    assert tokenize([0, 1, 0x90, 2, 3, 0xf8, 6]) == [[0x90, 2, 3], [0xf8]]
+
+
+def test_running_status():
+    """The last known status byte should be reused if omitted in a message."""
+    assert tokenize([0x9a, 3, 127, 3, 64, 3, 0, 0xb0, 2, 3, 4, 5]) == [
+        [0x9a, 3, 127], [0x9a, 3, 64], [0x9a, 3, 0],
+        [0xb0, 2, 3], [0xb0, 4, 5]
+    ]


Jajcus:
I was quite surprised finding out that MIDO cannot handle running status. Such a basic MIDI feature…

I am writing a new back-end and was feeding the parser with midi bytes using running status and wondered why only the first event is accepted.

m-vo:
@Jajcus did you test if the fixes from this PR work for you? Might be good to get some feedback.

Jajcus:
@m-vo No, I have not. I wanted to make my code work with the released version of mido, so I implemented a workaround even before I found this merge request.

rdoursenaud:
Per the official midi.org specification, page 5, running status is "For Voice and Mode messages only" aka Channel Messages.
System messages (Common, Real-Time and Exclusive) are not Running Status enabled and should exit running status mode.
Same goes for undefined Status bytes.

I don’t see such filtering in your code.

rdoursenaud:
I made a mistake in my previous comment. Real-Time messages are a special case and the only ones that should not reset/exit the Running status.

m-vo:
> Real-Time messages are a special case and the only ones that should not reset/exit the Running status.

The code only runs in `_feed_data_byte`, so IMHO this should not affect SysEx and other messages at all. We could in theory reset `self._last_status` in certain cases, but I didn't find anything in the spec about this. Do you see any downsides in the current implementation that we should address?

rdoursenaud:
The specs says the reverse.
Quote from MIDI 1.0 specification page 5 (10 in PDF form): "For Voice and Mode messages only" and "Running Status will be stopped when any other Status byte intervenes."

Common messages can have data bytes (F1, F2, F3) and SysEx has data bytes but are *not* Running Status enabled.

How does this test perform?
```py
def test_running_status_common():
    """The last known status byte should not be reused with common or exclusive message"""
    assert tokenize([0xf3, 1, 2]) != [
        [0xf3, 1], [0xf3, 2]
    ]
```
To be spec compliant it should pass and from just reading the code, I believe it fails.
The second data byte should be considered invalid and not yield another Song Select message.

FWIW I have a spec compliant Running Status implementation part of a work in progress to fix #363. I can extract/PR it since it seems to work fine while I continue working on the EOX issue which is more hairy.

m-vo:
Thanks for the explanation. I must have jumped over the "Running Status will be stopped …" part each time. :see_no_evil: 

> FWIW I have a spec compliant Running Status implementation part of a work in progress to fix https://github.com/mido/mido/issues/363. I can extract/PR it since it seems to work fine while I continue working on the EOX issue which is more hairy.

Sounds good. :+1:  Closing this PR in favor of #363 then.

rdoursenaud:
If anyone is willing to review my code, I just pushed my work in progress: https://github.com/EMATech/mido/tree/eox

The relevant running status implementation is there: https://github.com/EMATech/mido/commit/c6e13e53853dc53d9bd750af077ea852f843967e#diff-657b1756ea262acda7e2095a923fc3f18e64343552164609e0b5badd451151b4

m-vo:
> If anyone is willing to review my code, I just pushed my work in progress: [https://github.com/EMATech/mido/tree/eox](https://github.com/EMATech/mido/tree/eox?rgh-link-date=2022-04-05T22%3A55%3A05Z)
> 
> The relevant running status implementation is there: https://github.com/EMATech/mido/commit/c6e13e53853dc53d9bd750af077ea852f843967e#diff-657b1756ea262acda7e2095a923fc3f18e64343552164609e0b5badd451151b4

Thanks. Please consider opening a (draft) PR, because IMHO it's way easier to review this way.

PR #342, opened 2021-10-10
Set socket file buffering to 0

m-vo:
With `None` I actually *do* get buffering with Python 3.8 under Linux. Does `None` maybe imply using some standard value?

If buffering is active, the `_is_readable` check will return false even though data could be read.


rdoursenaud:
Looking at the implementation code, `None` means "Use the default buffering" and a `0` is indeed required to disable it.

PR #361, opened 2022-02-17
New teVirtualMidi backend

davr:
teVirtualMidi allows for creating virtual midi ports on windows. It's the same driver used by the loopMidi tool; the reason for using this backend instead of just using loopMidi, is that it allows for creating "non-loopback" midi ports. Where the input and the output don't just echo each other. This is needed for my very specific use case of writing a python script that sits as a translator between host software and a midi device, if I just used a loopback port, when the software wrote to the device, it would see it's own messages echoed back, which caused problems.

For this to work requires loopMidi being installed to get the driver. But don't actually need to run loopMidi
https://www.tobias-erichsen.de/software/loopmidi.html

### DIFF
--- a/mido/backends/tevirtualmidi.py
+++ b/mido/backends/tevirtualmidi.py
@@ -1,0 +1,110 @@
+"""
+Virtual port
+"""
+
+from __future__ import absolute_import
+from ..ports import BaseInput, BaseOutput
+import ctypes
+from ctypes import *
+
+MAX_MSG_SIZE = 1024
+
+vm = ctypes.CDLL("c:/windows/system32/teVirtualMIDI.dll")
+
+PARSE_RX = (1)
+PARSE_TX = (2)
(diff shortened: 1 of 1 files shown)


rdoursenaud:
Nice!
Any reason for not using https://github.com/complynx/pytemidi?

rdoursenaud:
I also forgot to mention that the SDK [is not free software](https://www.tobias-erichsen.de/software/virtualmidi/virtualmidi-sdk.html) and can’t be linked to MIDO unless authorized by the author.

davr:
1. Because I didn't think to see if someone had already done this? lol oh well, at least I got some learning experience with ctypes
2. This doesn't link to or use the SDK in any way, doesn't require the SDK to be installed. Just requires that the end user has installed loopMidi themselves. It does reference function names in the DLL which are also in the SDK, but I think that's acceptable (I'm not a lawyer etc)

PR #370, opened 2022-04-13
[WORK IN PROGRESS] Separate EOX message from SYSEX message

### LINKED ISSUE #241: keeping time information in rtmidi _callback_wrapper
Hi, 

would it be possible to modify the line 166 from 

 ```msg = Message.from_bytes(msg_data[0])```

to 

 ```msg = Message.from_bytes(msg_data[0], time=msg_data[1])```

I would then be able to use the time information.

### LINKED ISSUE #340: Support running status
Currently, the `Tokenizer` just drops data bytes if there was no status byte set in the current message:

https://github.com/mido/mido/blob/b884bc3c41d64da77bf6f3c879f823626c188e89/mido/tokenizer.py#L56-L65

The MIDI spec, however, allows to omit status bytes (see [MIDI running status](http://midi.teragonaudio.com/tech/midispec/run.htm)) in which case the last transmitted status should be used. In my case a hardware device uses this feature and by this totally scrambles the parser state. :cry: 

If I see this correctly, the only change needed would be to store the last status byte when completing a message with at least one data byte and only ignore 'stray data bytes' if nothing was set yet.

### LINKED ISSUE #363: Requiring EOX for creating a SysEx message is not spec compliant
### Specification

The MIDI specification states, page 5 (emphasis added by me):
"Exclusive messages [...] can be terminated either by an End of Exclusive (EOX) **or any other Status byte (except Real Time messages)**."

### Rationale

Currently, the parser does not interpret these SysEx messages and drops them!
It is also impossible to create a SysEx message without EOX from hex or string.

IMHO Since EOX (`0xF7`) is a Status byte, it should *not* be part of the SysEx message.

My understanding of "An EOX should always be sent at the end of a System Exclusive message" in the specification is that the sender is responsible for adding the EOX after a SysEx message, not that it is part of it!

I’d be happy to contribute a patch if you approve of these *potentially breaking* changes.

### Examples

```py
mido.parse_all(bytes.fromhex("f000801212"))
```
Current result: `[Message('note_off', channel=0, note=18, velocity=18, time=0)]`
Expected result: `[Message('sysex', data=(0,), time=0), Message('note_off', channel=0, note=18, velocity=18, time=0)]`

```py
mido.Message.fromhex("f000")
```
Current result: `ValueError: invalid sysex end byte 0`
Expected result: `Message('sysex', data=(0,), time=0)`

rdoursenaud:
NOT PRODUCTION READY, CONTAINS A BUNCH OF UNRESOLVED FIXMEs AND TODOs.
ONLY PUBLISHED FOR REVIEW PURPOSES!

Fixes #363

Also implements:
 - receiving Running Status (Does *not* implement sending with Running Status yet)
   Fixes #340
 - getting delta time from rtmidi backend (In a bad way for the moment)
   Fixes #241

Known bugs:
- [ ] Sending EOX alone seems to not be supported by the WinMM API and triggers an RtMidi error
- [ ] The argument used for RtMidi delta time is not fit for this purpose

TODO:
- [ ] Running status support:
    - [x] Receiving. See #341
    - [ ] Emitting
- [x] Get delta time from rtmidi backend
  - [ ] FIXME: put in a better argument. See #193
- [ ] EOX message support:
  - [x] Receiving as separate message
  - [ ] Always emit after SysEx if missing
- [ ] Optional SysEx filtering
  - [ ] Only parse messages compliant with a user provided lists of manufacturer ID codes.
        Pay special attention to the Universal System Exclusive IDs (0x7e & 0x7f) that should always be received
- [ ] Optionally return messages and/or raise exceptions on Undefined Status, Unexpected data or any Invalid message


rdoursenaud:
This tries to do too much in one go. I'll split every feature into their own commit for inclusion in mido version 2.
Work already started. See: https://github.com/mido/mido/tree/2.x

PR #383, opened 2022-08-22
put back necessary import in portmidi backend

brunoenten:
Reverse last commit by putting back byref import


rdoursenaud:
Thank you for your contribution.
I have cooked what I believe to be a better fix and will merge it soon.

PR #408, opened 2023-01-15
Removed 2.7 support. Now requires 3.7 or higher.

olemb:


olemb:
@bast, can you approve this?

.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [2.7, 3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
```
rdoursenaud, line 15: I suggest we add the current stable 3.10 and 3.11 versions.

rdoursenaud (changes requested):

.github/workflows/test.yml
olemb: Good point. Added these.

Tests fail now. Why is it trying to test with Python 3.1?

```
Error: Version 3.1 with arch x64 not found
```

Hmm.

olemb: Here's the list in .github/workflows/test.yml:

```
        python-version: [3.7, 3.8, 3.9, 3.10, 3.11]
```

No 3.1 there. Any ideas?

rdoursenaud: I think you need to quote the versions else they are interpreted as floats:
```
    python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
```

rdoursenaud: An as luck would have it, the versions we had were just casted correctly until now ;)

rdoursenaud:
I just found out you also missed two occurrences of `Python 2` in the docs:
- `messages.rst` line `103`
- `frozen_messages.rst` line `48`

.github/workflows/test.yml
olemb: Aha! I didn't consider that. Seems to work now. Thanks!

Also didn't think of looking for "Python 2" in the docs. Fixed those two. So nice to not have to proved two examples anymore.

bast (approved):

PR #413, opened 2023-01-19
Fix PortMidi backend

### LINKED ISSUE #372: 'byref' Error when opening a port with PortMidi
Hi, I'm getting the error:

`AttributeError: module 'mido.backends.portmidi_init' has no attribute 'byref'`

when trying to open an output port with the portmidi backend. *(I'm using mido version 1.2.10)*

I fixed it by reverting *backends/portmidi_init.py* back to commit `bdc9249`

just a little heads up.

rdoursenaud:
Hopefully for good this time using the library recommended method of importing everything.

This issue has already crept up several times:
- d051be32e4f6ee52f4570bcfed445aae20a92526
- bdc924960160e2677dd196953d622c89b58cbc5d
- #120
- #383

Probably because the `byref` import was marked unused by IDEs and linters.

Fixes #372
Closes #383

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,3 +2,6 @@
 ignore = F401, C901, F901
 exclude = .git,__pycache__,docs/source/conf.py,old,build,dist
 max-complexity = 10
+per-file-ignores =
+    # We really need to use an `import *` here to stop confusing everyone.
+    mido/backends/portmidi_init.py:F403, F405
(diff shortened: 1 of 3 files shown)


akx:
I don't think the `import *` is a good idea – just explicitly export the symbols you want to export from the backend, or alternately import `ctypes.byref` in `portmidi.py`...

PR #451, opened 2023-01-23
Cleaned up the logo

rdoursenaud:
Used the Wikimedia Commons DIN-5 Diagram as a template: https://commons.wikimedia.org/wiki/File:DIN-5_Diagram.svg

Borrowed Python gradient colors.

Added a shadow effect.


bast:
I really like the new look. I am unsure, however, whether we should give up the square-ness of it since square-ish logos are convenient for avatars and social media icons/avatars which are typically square or round but not rectangular/oblong.

rdoursenaud:
You're right, I didn't think of these use cases!
I can add a square/circle-compatible variant by rendering without the drop shadow.
I'll give it a go next week.

bast:
Thanks!

rdoursenaud:
A bit later than expected but done!
There is now a square, round compatible PNG export.
I documented my process in the readme.

I was also thinking about adding horizontal and vertical versions with `mido` spelled out.
I have no idea which font to use. The Python logo font is not free: [Flux](https://www.myfonts.com/collections/flux-font-t-26). The old midi logo's font was cool but has been replaced by something that, to me, feels very generic and soulless.
Any suggestions?

bast:
Thanks a lot! I will review and comment very soon. Thank you for your work on this.

bast (approved):
Thanks! I really like these.

PR #470, opened 2023-03-29
move complex function out of __iter__() first loop

Frnot:
The MidiFile.iter() function calls merge_tracks() on its first iteration. I have examples of this function taking many tenths of seconds to execute. This causes issues when using Midifile.play() to provide the timing for midi playback, as the first note(s) will play long after their calculated 'input_time' (because they have to wait on merge_tracks()).

This change calls merge_tracks on MidiFile instantiation and after each add_track() call, removing the need to call it in MidiFile.iter()

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -315,6 +315,9 @@ class MidiFile:
         elif self.filename is not None:
             with open(filename, 'rb') as file:
                 self._load(file)
+        
+        if self.type != 2:
+            self.merged_track = merge_tracks(self.tracks)
 
     def add_track(self, name=None):
         """Add a new track to the file.
@@ -326,6 +329,8 @@ class MidiFile:
         if name is not None:
             track.name = name
         self.tracks.append(track)
+        if self.type != 2:
+            self.merged_track = merge_tracks(self.tracks)
         return track
 
     def _load(self, infile):
@@ -374,7 +379,7 @@ class MidiFile:
             raise TypeError("can't merge tracks in type 2 (asynchronous) file")
 
         tempo = DEFAULT_TEMPO
-        for msg in merge_tracks(self.tracks):
+        for msg in self.merged_track:
             # Convert message time from absolute time
             # in ticks to relative time in seconds.
             if msg.time > 0:
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -315,7 +315,7 @@ class MidiFile:
         elif self.filename is not None:
             with open(filename, 'rb') as file:
                 self._load(file)
-        
+        # merge tracks at load time to prevent timing error on first call to __iter__()
         if self.type != 2:
             self.merged_track = merge_tracks(self.tracks)
 
@@ -329,6 +329,7 @@ class MidiFile:
         if name is not None:
             track.name = name
         self.tracks.append(track)
+        # merge new track immediately to prevent timing error on first call to __iter__()
         if self.type != 2:
             self.merged_track = merge_tracks(self.tracks)
         return track
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -315,7 +315,8 @@ class MidiFile:
         elif self.filename is not None:
             with open(filename, 'rb') as file:
                 self._load(file)
-        # merge tracks at load time to prevent timing error on first call to __iter__()
+        # merge tracks at load time to prevent timing error on 
+        # first call to __iter__()
         if self.type != 2:
             self.merged_track = merge_tracks(self.tracks)
 
@@ -329,7 +330,8 @@ class MidiFile:
         if name is not None:
             track.name = name
         self.tracks.append(track)
-        # merge new track immediately to prevent timing error on first call to __iter__()
+        # merge new track immediately to prevent timing error on 
+        # first call to __iter__()
         if self.type != 2:
             self.merged_track = merge_tracks(self.tracks)
         return track
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -315,7 +315,7 @@ class MidiFile:
         elif self.filename is not None:
             with open(filename, 'rb') as file:
                 self._load(file)
-        # merge tracks at load time to prevent timing error on 
+        # merge tracks at load time to prevent timing error on
         # first call to __iter__()
         if self.type != 2:
             self.merged_track = merge_tracks(self.tracks)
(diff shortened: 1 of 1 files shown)


almostimplemented:
Interesting find. I've been using `mido` for playback but none of my MIDI files have multiple tracks. And furthermore, they all have their first note onset after 1-2 seconds. So the delay has never caused an issue with my playback system (thanks to the fact that we compute the time to `sleep` based on the current time and not just blindly `sleep`ing by relative time values).

I think the main bottleneck in `merge_tracks` is the sort.

I wonder if this would be a more elegant solution:

```
import heapq 

...

def merge_tracks(tracks):
    return fix_end_of_track(
        _to_reltime(
            heapq.merge(
                *[_to_abstime(track) for track in tracks], key=lambda msg: msg.time
            )
        )
    )
```

This takes advantage of the fact that the individual `tracks` are already sorted, so we can just use a min-heap with the absolute time values as keys.

Frnot:
While I think improving the time complexity of merge_tracks() is a good idea, I still think it should be removed from __iter__() to prevent redundant computations every time the midifile is iterated through.

We could memoize merge_tracks() to perform the sort once only when needed. However, we still risk the priority queue implementation being too slow to execute on extreme cases. I'd have to investigate if these extreme cases are anyway realistic though.

almostimplemented:
I am really glad you mentioned that. 

I decided to profile the `merge_tracks` method on a large synthetic multi-track MIDI file, and it looks like sorting is definitely not the issue. The bottleneck is `checks.py`.

```python
import mido
import time
from random import randint

mf = mido.MidiFile()

for _ in range(5):
    t = mido.MidiTrack()
    for _ in range(10000):
        msg = mido.Message("note_on", note=72, time=randint(10, 1000))
        msg = mido.Message("note_off", note=72, time=randint(10, 1000))
        t.append(msg)
    mf.tracks.append(t)

now = time.time()
i = 0
for _ in mido.merge_tracks(mf.tracks):
    i += 1

print(i)
print(time.time() - now)
```
Using `cProfile`, sorting by total time:
```
(fresh) ➜  examples git:(main) ✗ python -O -m cProfile -s tottime benchmark.py
50001
0.9684998989105225
         10965496 function calls (10965197 primitive calls) in 1.459 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   300000    0.215    0.000    1.088    0.000 checks.py:89(check_msgdict)
  1500000    0.191    0.000    0.850    0.000 checks.py:85(check_value)
  1200001    0.133    0.000    0.314    0.000 {built-in method _abc._abc_instancecheck}
  1200930    0.122    0.000    0.520    0.000 {built-in method builtins.isinstance}
900006/900004    0.117    0.000    0.117    0.000 {built-in method _abc._abc_subclasscheck}
  1200001    0.084    0.000    0.398    0.000 abc.py:117(__instancecheck__)
   200000    0.075    0.000    0.920    0.000 messages.py:110(__init__)
   100000    0.072    0.000    0.922    0.000 messages.py:117(copy)
900006/900004    0.064    0.000    0.181    0.000 abc.py:121(__subclasscheck__)
...
        6    0.002    0.000    0.004    0.001 {method 'sort' of 'list' objects}
```
And sorting by cumulative time:
```
(fresh) ➜  examples git:(main) ✗ python -O -m cProfile -s cumtime benchmark.py
50001
0.9511542320251465
         10965496 function calls (10965197 primitive calls) in 1.439 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     33/1    0.000    0.000    1.439    1.439 {built-in method builtins.exec}
        1    0.037    0.037    1.439    1.439 benchmark.py:1(<module>)
   300000    0.214    0.000    1.073    0.000 checks.py:89(check_msgdict)
        1    0.003    0.003    0.946    0.946 tracks.py:102(merge_tracks)
   200000    0.075    0.000    0.910    0.000 messages.py:110(__init__)
   100000    0.071    0.000    0.905    0.000 messages.py:117(copy)
  1500000    0.179    0.000    0.836    0.000 checks.py:85(check_value)
  1200930    0.121    0.000    0.519    0.000 {built-in method builtins.isinstance}
    50002    0.006    0.000    0.476    0.000 tracks.py:80(fix_end_of_track)
```

Frnot:
That makes sense. I suppose that the checks preformed are redundant in this case, since we are combining tracks that have already been checked. However, rewriting the library to prevent the redundant checks doesn't seem worth the effort. I still maintain that my original suggestion (the content of this PR) solves the immediate problem most elegantly.

almostimplemented:
@Frnot I'm not arguing against moving the `merge_tracks`, or at least providing a way to eagerly compute it. There is no need to look at these changes as mutually exclusive.

Frnot:
For sure, but in the spirit of getting this pull request merged (are the maintainers even still active?) I am attempting to keep the scope limited

almostimplemented:
Point taken -- I will open a new issue to improve the performance of `merge_tracks` so as to not distract from your direct need.

mido/midifiles/midifiles.py
```diff
@@ -315,6 +315,9 @@ def __init__(self, filename=None, file=None,
         elif self.filename is not None:
             with open(filename, 'rb') as file:
                 self._load(file)
+
```
rdoursenaud, line 318: Please add a small comment to explain the rationale behind merging at load time.

mido/midifiles/midifiles.py
```diff
@@ -326,6 +329,8 @@ def add_track(self, name=None):
         if name is not None:
             track.name = name
         self.tracks.append(track)
+        if self.type != 2:
```
rdoursenaud, line 332: Same.

rdoursenaud:
Thanks for hunting this down and for the nice insights!

rdoursenaud (approved):

Frnot:
I made the changes and committed them. I don't know how github works, so I don't know if/how I could make the changes in the proposal comment windows.

rdoursenaud:
It's ok, the changes are automatically propagated as long as you use the same branch. I've already approved them. Thank you again!

rdoursenaud (approved):

Frnot:
@rdoursenaud I ran pytest, flake8 and check_manifest locally and corrected any errors. I think the documentation at https://mido.readthedocs.io/en/latest/contributing.html needs to be updated. I had to check the github workflow to find the right commands to run.

Frnot:
@rdoursenaud is there any estimate for when this change will be merged?

rdoursenaud (approved):

rdoursenaud:
> I think the documentation at https://mido.readthedocs.io/en/latest/contributing.html needs to be updated. I had to check the github workflow to find the right commands to run.

Sorry for that. I don't have permissions on the ReadTheDocs project yet so I'm unfortunately not able to update it at the moment. You did well checking the GitHub workflow file!

> is there any estimate for when this change will be merged?

Yes, right now :)

rdoursenaud:
Thank you very much for your contribution!

stevenjoezhang:
Hello, this change may potentially cause some minor issues. For instance, when a track is modified, such as when a new message is inserted, `self.merged_track` will not be updated. This could lead to situations that might confuse users.

PR #483, opened 2023-05-23
Don't use star imports in portmidi_init

akx:
Follows up on #413.

Instead of using star imports, let's just be specific about what we're using from `ctypes` and that's that 😄


rdoursenaud (approved):
Thanks, your changes makes sense.
I didn't go deep enough with my quick and dirty fix.
Great contribution!

PR #488, opened 2023-07-04
Separate GitHub workflow test steps

rdoursenaud:
Hopefully allowing better readability of output.


rdoursenaud:
All tests should now pass.

To explain what I did and why:
- Running all the tests under the same `run` directive doesn't take into account all return codes.
  This was hiding some already failing tests and preventing from seeing what was really failing.
  So I separated each test into its own `run` directive.
- Doing so revealed some missing license files that I've added
- It also revealed that `check-manifest` errors under windows and python < 3.10 with a `WinError 2`.
  Since I haven't been able to find the root cause, I have excluded tests for python versions < 3.10 under windows to avoid false negatives.
  If someone knows better about this issue, please propose a pull request.

PR #523, opened 2023-07-28
fix portmidi backend sysex

tweakoz:
Fix for error:
AttributeError: module 'mido.backends.portmidi_init' has no attribute 'c_char_p'

### DIFF
--- a/mido/backends/portmidi.py
+++ b/mido/backends/portmidi.py
@@ -262,7 +262,7 @@ class Output(PortCommon, BaseOutput):
     def _send(self, message):
         if message.type == 'sysex':
             # Sysex messages are written as a string.
-            string = pm.c_char_p(bytes(message.bin()))
+            string = pm.ctypes.c_char_p(bytes(message.bin()))
             timestamp = 0  # Ignored when latency = 0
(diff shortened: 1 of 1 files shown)


mido/backends/portmidi.py
```diff
@@ -262,7 +262,7 @@ class Output(PortCommon, BaseOutput):
     def _send(self, message):
         if message.type == 'sysex':
             # Sysex messages are written as a string.
-            string = pm.c_char_p(bytes(message.bin()))
+            string = pm.ctypes.c_char_p(bytes(message.bin()))
```
rdoursenaud, line 265: I believe it should be `ctypes.c_char_p` instead of `pm.ctypes.c_char_p` since `ctypes` is available in the context.

rdoursenaud (changes requested):
Thanks for the fix.
Please see my comment in the review.

mido/backends/portmidi.py
tweakoz: change applied.

rdoursenaud (approved):
Thanks!

Issue #528, opened 2023-08-02
Avoid numerical errors by using `fractions` where appropriate


page200:
The time in seconds is computed by dividing precise numbers, so it should be stored precisely by using [`fractions`](https://docs.python.org/3/library/fractions.html). Otherwise numerical errors can accumulate as the time intervals are being summed, which can affect people who leave their equipment (including art installations) turned on for long periods of time or some algorithms. The errors can accumulate even faster if there are MIDI messages every tick (for example pitch-bend messages that realize a vibrato effect). And even if the numerical error is tiny, there's a chance that it will get amplified for example by discretization. For example the time discretization of "piano rolls" in general isn't in sync with MIDI ticks. With such errors, `mido` can't be used together with libraries that compute the time correctly.

This bugfix seems to be quite non-breaking because it might affect only `delta_seconds` (not yet implemented under that name, see issue #527) and maybe `fractions` is quite compatible with how people use `mido`.

Issue #457, opened 2023-01-31
Add support for SMPTE based SMF division


rdoursenaud:
Standard MIDI Files (SMF) can specify their time base division in two formats:
- [x] ticks per quarter-note (aka notation-based)
- [ ] negative SMPTE format + ticks per frame (aka time-based)

Currently, `mido` only supports the first format and creates invalid data when the second is present.

This information is stored in the `Header Chunk` of the SMF in the `division` field:
```py
b'MThd'    # Header chunk magic (4 bytes)
<length>   # Length of the following data. At least 6 bytes. (1 byte)
<ff>       # Format type (2 bytes representing a 16-bit word MSB first aka big-endian)
<ntrks>    # Number of tracks (2 bytes representing a 16-bit word MSB first aka big-endian)
<division> # Time base division (2 bytes representing a 16-bit word MSB first aka big-endian)
```

The data is a 16-bit word (2 bytes big-endian).
The most significat bit represents the time base division format:
- 0: ticks-per quarter note.
  The value is stored in the 15 other bits in big-endian form.
- 1: SMPTE.
  Negative SMPTE format encodes the framerate onto the remaining 7 bits of the most significant byte.
  The least significant byte encodes the `tick per frames`.

Challenges:
`mido` always expects `ticks_per_beat` (sic) for the moment.
A thorough review of the implications is required before implementation.
Including, but not limited to:
- [ ] Tempo determination
- [ ] Unit conversions
- [ ] Message time property

PR #531, opened 2023-08-04
Improve conversions between ticks and seconds

### LINKED ISSUE #526: Wrong code comments about relative and absolute time?
Is my understanding correct that "absolute time" can be interpreted as the time that has passed since the beginning of a song, whereas "relative time" can be interpreted as the time that has passed since the previous MIDI message?

If so, then `tick2second()` and `second2tick()` are capable of converting _absolute_ time in ticks or seconds to _absolute_ time in seconds or ticks, as well as converting _relative_ time in ticks or seconds to _relative_ time in seconds or ticks. And they are often =used for the latter (converting _relative_ time).

Therefore, the four occurrences of "absolute" should be removed from their docstrings.

And then [this comment](https://github.com/mido/mido/blob/8e5164925f720c79d1e2a7c0e0d0ae13fcfdb7ab/mido/midifiles/midifiles.py#L390-L391) is wrong: _relative_ (not _absolute_) time in ticks is being converted to _relative_ time in seconds there.

By the way, why the singular "tick", "second" rather than "ticks", "seconds" in the name of the functions `tick2second()` and `second2tick()` and their input argument `tick`?

rdoursenaud:
Deprecates tick2second in favor of ticks2seconds and second2tick in favor of seconds2ticks.
Returns fractional time is seconds per #528
Removes erroneous references to absolute in docstrings and comments.

Fixes #526

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: CC0-1.0
 
 [flake8]
-ignore = F401, C901, F901
+# W503 and W504 are incompatible. Prefer the modern approach.
+ignore = F401, C901, F901, W503
 exclude = .git,__pycache__,docs/conf.py,old,build,dist
 max-complexity = 10
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -181,7 +181,11 @@ Standard MIDI Files
 
 .. autofunction:: tick2second
 
+.. autofunction:: ticks2seconds
+
 .. autofunction:: second2tick
+
+.. autofunction:: seconds2ticks
 
 .. autofunction:: bpm2tempo
 
--- a/docs/files/midi.rst
+++ b/docs/files/midi.rst
@@ -237,11 +237,13 @@ in number of beats per minute (BPM) or microseconds per quarter note, see
 `MIDI Tempo vs. BPM`_ above) and ticks per per quarter note have to be decided
 upon.
 
-You can use :py:func:`tick2second` and :py:func:`second2tick` to convert to
+You can use :py:func:`ticks2seconds` and :py:func:`seconds2ticks` to convert to
 and from seconds and ticks. Note that integer rounding of the result might be
 necessary because MIDI files require ticks to be integers.
 
-If you have a lot of rounding errors you should increase the time resolution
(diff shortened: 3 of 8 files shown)


page200:
Thank you for the nice discussions and commits! :)

One comment: I'd favor renaming `resolution` to the units it is given in, such as `ticks_per_quarternote`.

rdoursenaud:
The unit is a property that is not descriptive enough IMHO.

I find that `resolution` conveys the intention better and I intend to use the same field name when revamping the `MIDIFile` class.

I pondered using `time_ticks`, `time_seconds`, `resolution_tpqn` and `tempo_uspqn` but I think it makes the API overly verbose and may confuse users more because of the lack of context provided by the documentation.

rdoursenaud:
And thanks for the review!

page200:
In what way do you consider `ticks_per_quarternote` not descriptive enough, I mean, what variables except for resolution can be given in ticks per quarternotes? If there are such variables, let's call this one `resolution_in_ticks_per_quarternote` or so. I suspect that `ticks_per_quarternote` is a good choice. Informative names are good.

Leaving out units would be a huge pain, especially in the context of MIDI, where time is so complicated already due to ticks vs. seconds vs. quarternotes and the various events that affect how to convert between those "units". "Resolution" might be given in ticks per quarternote, or ticks per second, or ticks per millisecond, or ticks per minute, or ticks per bar. Or maybe other things can also be considered "resolution", for example something related to the limited numerical precision of the inexact number of seconds that users might assume or recall from before the fix of #528, or the limited numerical precision of data in MIDI events such as tempo changes, or an assumption that `mido` voluntarily chooses to limit the resolution to some value by rounding. All these units, conversions, MIDI events (including how they affect conversions over time and how they have limited precision) and optional roundings are dauting, and `ticks_per_quarternote` nails it, it seems.

rdoursenaud:
`ticks_per_quarternote` is just a unit it doesn't convey its _purpose_.
I just checked the SMF specification and they call this `division`, maybe we should use this to avoid bikeshedding as it is the trusted source.

rdoursenaud:
Also the `division` field will need to a unit anyway since it can either be expressed in ticks per quarter note or ticks per frame.
We don't have support for the latter at the moment. See #457.

Issue #530, opened 2023-08-04
"Pass by value" vs. "pass by reference" inconsistent in `fix_end_of_track()`


page200:
`fix_end_of_track()` yields `msg` or `msg.copy()` depending on whether the previous message was a premature `'end_of_track'` message.

If I understand the difference between `msg` and `msg.copy()` correctly, the user of `fix_end_of_track()`, when trying to modify messages (for example changing the note from A to A♯) yielded by `fix_end_of_track()`, will also modify some of the notes of the original input object of `fix_end_of_track()` but not other notes. So the changes of that original object will be inconsistent, they will depend on whether the previous message was `'end_of_track'`.

We can either replace `yield msg` by `yield msg.copy()` or remove `if accum:` and `else: yield msg`.

rdoursenaud:
I don't understand your description very well but I think this is coming from `Message` objects being mutable. IMHO `Message` objects coming from either `Ports` (MIDI Inputs) or `Tracks` (MIDI Files) should be immutable. Only objects the user creates should be mutable.

page200:
Yes, this is about `Message` objects being mutable. Even if we make some of them immutable, the bug will still affect mutable ones. Even if `mido` didn't create any mutable ones, users' custom code or future `mido` versions might create mutable ones for some reason and the bug would resurface if not fixed.

rdoursenaud:
Would freezing messages when associating them to a `MIDIFile` fix this?

page200:
Even if we make it impossible for users to pass mutable "messages" as inputs to `fix_end_of_track()` (which we shouldn't), users might still modify the code when experimenting with some algorithms or improving `mido`, and then this bug would affect them.

Consistently using `msg.copy()` instead of mixing `msg` with `msg.copy()` uses more memory though (unless we use "copy on write"). So here's a new idea:

Let's have an additional input argument to `fix_end_of_track()` that decides whether the safe (default for users) or memory-friendly algorithm should be used. Then `merge_tracks()` and `MidiFile.save()` can use the memory-friendly one, and users can use the safe one by default or whatever they choose. Would you welcome this as a pull request?

rdoursenaud:
Since this is part of the public-ish API (not prepended by `_` but not documented either), I'd accept such a PR, even though it clearly hasn't been designed for direct user interaction, rather as a facility to ensure that we don't save non-compliant files.

page200:
(Note to myself: `merge_tracks()` has to use the safe mode because its output can be saved.)

PR #532, opened 2023-08-04
Fix issue #530

page200:
The rarely used function `fix_end_of_track()` now has the input argument `readonly_promise`, default `False`. `True` saves memory but yields outputs that should not be modified because their modification would affect the original inputs in weird incoherent ways. `merge_tracks()` uses `readonly_promise=False` because it stores the outputs of `fix_end_of_track()` and thus they might get modified later. `MidiFile.save()` uses `readonly_promise=True` because it discards the outputs of `fix_end_of_track()` without modifying them. Those are deliberate choices and thus even the usage of the default parameter `readonly_promise=False` is explicitly coded.

This should fix issue #530.

Attention: Tests were not run and not adapted.

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -241,7 +241,7 @@ def write_track(outfile, track):
     data = bytearray()
 
     running_status_byte = None
-    for msg in fix_end_of_track(track, readonly_promise=True):
+    for msg in fix_end_of_track(track, safe=False):
         if not isinstance(msg.time, Integral):
             raise ValueError('message time must be int in MIDI file')
         if msg.time < 0:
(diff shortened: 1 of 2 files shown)


rdoursenaud:
Thanks.

Please fix the linter issues so we can run the tests.

I kinda preferred the `safe` keyword. `readonly_promise` is a bit verbose and needs knowledge of the internals to be properly understood. Description of the internal mechanisms can be added to the docstring to clarify the intent.

page200:
I've fixed the linter issues and renamed the argument back to `safe`. It passes the tests: https://github.com/page200/mido/actions/runs/5775408524

page200:
I just had merely renamed the branch. Apparently that automatically closes the PR and says that the branch was deleted. So I renamed it back and I'm reopening the PR.

rdoursenaud (approved):
Looks good to me. Thanks!

Issue #527, opened 2023-08-02
Clearer name for `time` attribute


page200:
The attribute name `time` doesn't explain whether it is for time since the beginning of the song or since the previous MIDI message, nor whether it is given in seconds or milliseconds or ticks.

Moreover, the units (ticks vs. seconds) are inconsistent, e.g. `print(mid)` and `print([*mid])` shows different time values, despite the attribute being called `time` in both cases.

Instead, there can be two attributes: `delta_ticks` and `delta_seconds`. "Delta" in the name indicates that these are the ticks/seconds since the previous MIDI message, rather than since the beginning of the song.

Converting between the two requires knowing the previous tempo-change message. That might be the reason why the time in seconds is computed only during a loop through all messages (i.e. during `__iter__()`) and then discarded. One option is to keep it that way. (An alternative is to keep track of tempo changes and let the setters of `delta_ticks` and `delta_seconds` set both attributes at once. As we look into keeping track of pitch-bend range changes in issue #514, insights about keeping track of tempo changes might emerge.)

page200:
`delta_seconds` should use `fractions`, see issue #528.

Issue #514, opened 2023-07-21
Pitch-bend range


page200:
In MIDI, the pitch-bend range is `2` semitones by default, but [can be modified via specific RPN messages, usually followed by specific "Data Entry" messages][1].

It would be nice to have functionality that understands those messages and can report the correct pitch or pitch-bend range.

As far as I understand, the pitch-bend range is set by three to four consecutive MIDI control messages. In hexadecimal, they look like this (with variables shown in **bold**):

B**0**6500
B**0**6400
B**0**06**02**
B**0**26**04**

The "**0**" after the "B" is the MIDI channel for which the pitch-bend range is being set.
The "**02**" in the end of the third message is the number of semitones by which the pitch can bend.
The "**04**" in the end of the fourth message is the number of [cents](https://en.wikipedia.org/wiki/Cent_(music)) to be added to the semitones.
The last message is optional, i.e. if there is no message of the form /B.26../ (like "B**0**26**04**" in the example) then the number of cents is zero.

  [1]: http://midi.teragonaudio.com/tech/midispec/rpn.htm

rdoursenaud:
This is one level of abstraction above what mido is currently providing.

Nothing prevents one from building a RPN/NRPN interpreter above mido though.

Don't get your hopes high unless there is a lot of interest in such a feature from the community.

page200:
Many mido-based repos intend to have correct pitch info. But the mistake of using the default pitch-bend range of 2 semitones without taking pitch-bend range messages into account is not obvious because not all MIDI files/streams have such messages and the correct pitch is not always known to the user. So they go through the tedious process of discovering this mistake late (after trying many MIDI files), searching for the reasons, implementing an RPN interpreter, and propagating the fix to their dependents and forks (of which there are many because the issue takes long to discover), often failing at one of those steps.

Do we want each such repo to go through this process separately, or would this functionality be welcome in mido?

There would be a lot of interest in the community if the community knew about this.

rdoursenaud:
Thank you for your insight.

To be clear, I'm not against anyone contributing & maintaining such feature.
I'd welcome it even!

It could belong in a `GM1 `(For General MIDI System Level 1) module since it's specified in there (Specification reference: **MMA-00007/RP-003**).
The default _±2 semitones_ range comes from this specification and was not part of the `MIDI 1.0` specification nor merged later.
[...]

PR #535, opened 2023-08-06
Pitch-bend range (and `delta_ticks`)

page200:
This pull request fixes issue #514. Specifically, `MidiFile.__iter__()` now keeps track of MIDI events (three to four specific consecutive events) that affect the pitch-bend range.

delta_seconds, current tempo (anytime _after_ `'set_tempo'` events), and GM1 stuff (pitch-bend range in semitones and pitch bend in semitones) depend on earlier MIDI events and are not merely part of the current MIDI event. Thus, they are reported via a separate dictionary named `inferred` (name seems slightly more future-proof/general than "`contextual`") instead of storing them inside the MIDI event object.

This dictionary is the ideal place for reporting the time in seconds. Thus, the time in the MIDI event can remain in ticks as per the MIDI specification instead of unexpectedly getting overwritten by seconds. (Regarding "unexpectedly", see a note in issue #527: `time` values in `print(mid)` and `print([*mid])` are different.) For `mido` users to not have to adapt their code twice, I'm taking this opportunity to also fix issue #527, i.e. rename `time` to `delta_ticks`.


rdoursenaud:
Thanks for your efforts, I however can't accept these changes as-is for the reasons outlined below:

### Pitch bend

pitch-bend related changes have no business inside the midi file decoding logic.

I'm not sure you understood what I said in https://github.com/mido/mido/issues/514#issuecomment-1646595870 the way I intended.

What I meant was _independent code_ in its own [module](https://docs.python.org/3/tutorial/modules.html) that developers can call _at their discretion_ for `General MIDI 1` decoding. This hypothetical `gm1` module could start by implementing pitch-bend range support but should be designed to be _extensible_ with other General MIDI 1 specifics.

To play nicely with `mido`, it should support being used with **protocol elements** (Namely, pre-decoded `RPN` messages, see below.) without any file context.
Only then can we provide facilities for helping decoding into context, like `MIDIFile` and `Port` functions _decorators_ for example.

Furthermore, before even going there, we'd have to implement a generic `RPN/NRPN` decoding system to base work upon since _every_ RPN/NRPN has the same basic decoding routines. We don't want boilerplate code repeated for each RPN/NRPN.

Let me know in #514 if you want me to help draft a suitable implementation design.

### Time

Finally, except when absolutely necessary, please refrain from mixing several subjects into the same pull request. Focused pull requests are easier to review and accept and makes life easier for everyone involved.

Here the changes related to `time` are bigger than those related to the PR title, both in size and scope.

Since the `time` property is part of mido's public API with thousand of users, it has a greater impact and can't be changed without proper deprecation warnings and documentation first.
This means keeping the property and alerting the user by raising a `DeprecationWarning` each time it's used explaining why it's been changed and directing him into choosing the relevant replacement for his use-case.
It also requires us to bump our major release number to stay SemVer compliant, hence why I set #527 milestone for version 2.0.0 where several API breaking changes are scheduled for.

### Conclusion

I'll keep the PR open for you to make appropriate changes it you wish to.
If you prefer, feel free to close it and eventually open another.
Thank you for your understanding.

page200:
> pitch-bend related changes have no business inside the midi file decoding logic.

Why does [`MidiFile.__iter__()`](https://github.com/mido/mido/blob/main/mido/midifiles/midifiles.py#L382-L400) keep track of tempo changes and compute time in seconds accordingly, but shouldn't keep track of pitch-bend range changes and compute pitch bend in semitones accordingly? If this is about separating General MIDI 1, see https://github.com/mido/mido/issues/514#issuecomment-1667826649

What kind of decoding of RPNs/NRPNs do you mean? Something like converting `0x6500` to `'pitchbend_range_message_1'`, `0x6400` to `'pitchbend_range_message_2'`, and `0x06` to `'pitchbend_range_message_3'`? Isn't my decoding (i.e. reporting the number of semitones) good?

rdoursenaud:
> Why does [`MidiFile.__iter__()`](https://github.com/mido/mido/blob/main/mido/midifiles/midifiles.py#L382-L400) keep track of tempo changes and compute time in seconds accordingly...

Ideally it shouldn't. It should be handled by a separate construct (`MIDIFilePlayer` maybe?). However, tempo information is **only** delivered in MIDI files in this form (Set Tempo Meta Event) and is required for proper interpretation so it makes sense to aggregate the functionality in this object for simplicity's sake.

> ...but shouldn't keep track of pitch-bend range changes and compute pitch bend in semitones accordingly?

Whereas pitch bend changes and the RPN pitch bend range are messages from the protocol. These do not only appear in files but also « on the wire » and we should be able to manipulate them like any other mido object.

> If this is about separating General MIDI 1, see [#514 (comment)](https://github.com/mido/mido/issues/514#issuecomment-1667826649). If on the other hand this is about being able to process data that didn't come from a MIDI file, then the code in `MidiFile.__iter__()` should be moved outside of `MidiFile` as well.

See above for my stance.

> What kind of decoding of RPNs/NRPNs do you mean? Something like converting `0x6500` to `'pitchbend_range_message_1'`, `0x6400` to `'pitchbend_range_message_2'`, and `0x06` to `'pitchbend_range_message_3'`? That can be changed later without affecting how the correct pitch bend in semitones gets reported from the user's perspective.

All Registered Parameter Numbers and Non-Registered Parameter Numbers works the same way. They have a 16 bit (2 bytes) identifier and are followed by either:
- Data Entry (1 or 2 bytes)
- Data Increment (1 byte)
- Data Decrement (1 byte)

They are transmitted using the appropriate control changes messages.
Usually LSB first but the specification states that a receiver should be able to interpret a single LSB or MSB. This means decoding is stateful (the system has to remember previous messages to correctly interpret the latter).

The pitch bend sensitivity (In MIDI 1.0) or pitch bend range (Refined In General MIDI 1) is the first Registered Parameter Number (RPN) with number `0x00 0x00`.

As you can see, decoding RPN/NRPN is not trivial.

Since mido is all about providing high level objects to interact with, I believe RPN and NRPN are good candidates for dedicated objects that can be constructed from a bunch of messages (built, received or read from a file), decoding them to their useful values (ID, name, value) and encoded to the proper messages sequence for transmission or storage.

This would make the tracking of their state trivial for the purpose you seem to chase: tracking pitch.

page200:
> pitch bend changes and the _RPN_ pitch bend range are messages from the protocol. These do **not** only appear in files but also « on the wire » and we should be able to manipulate them like any other mido object.

Shouldn't we be able to also manipulate _Set Tempo_ Meta Events?

How is the timing of events "on the wire" with `delta_ticks > 0` interpreted if tempo is unknown? Or is `delta_ticks` always zero "on the wire"?

_RPN_ and _NRPN_ may or may not have dedicated objects. Pitch bend in semitones may or may not be reported to the user. There are all four combinations. We can start reporting the pitch to the user (with something like this pull request or in some other way) and introduce objects for _RPN_ and _NRPN_ anytime later. Making both changes at once would be slightly less work, but the work of the pitch reporting is already done (and is probably easy to further polish, if necessary), and adding objects for _RPN_ and _NRPN_ will probably be less work than doing everything from scratch. Moreover, adding objects for _RPN_ and _NRPN_ shouldn't delay the access to pitch info.

PR #552, opened 2023-09-25
Allow channel_prefix without data

mrexodia:
Files like this are generated by KlavarScript (https://klavar.com/en/downloads-en). Most likely this is a bug in the program, because the MIDI spec says the sequence should be `FF 20 01 cc` but with these changes my files load fine!

image

### DIFF
--- a/mido/midifiles/meta.py
+++ b/mido/midifiles/meta.py
@@ -256,7 +256,9 @@ class MetaSpec_channel_prefix(MetaSpec):
     defaults = [0]
 
     def decode(self, message, data):
-        message.channel = data[0]
+        # Message with length 0 can occur in some files.
+        if len(data) > 0:
+            message.channel = data[0]
 
     def encode(self, message):
         return [message.channel]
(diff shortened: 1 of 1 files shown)


rdoursenaud:
Hi and thanks for proposing a fix.

Could you please share an offending file or at the very least a console output when loading the file with `debug=True`:
```py
import mido

mido.MidiFile('file.mid', debug=True)
```

Also, have you notified the author of KlavarScript about the issue?
While fixing it here seems trivial, it's often better to fix the root cause.

Finally, I'm in the process of rewriting mido's Standard MIDI File parser and will ignore (with a warning) such malformed events as per specification.
Is this acceptable for you or would you like the option to access those events for further processing?

Thanks!

mrexodia:
Yeah I’ll provide the file. The author of the tool hasn’t been responsive since 2014 unfortunately.

In this case the event is correct, the tool itself shows that it’s a “midi channel 0. I think my patch should work fine and the event shouldn’t be dropped…

mrexodia:
[bug.zip](https://github.com/mido/mido/files/13260832/bug.zip)

Sorry for the delay. I rebased my branch and attached the non-confirming file. In general I think my patch is quite harmless, since the packet length is stored earlier so it won't cause a domino effect of corruption.

mrexodia:
Rebased this again, any chance for a merge?

Issue #537, opened 2023-08-08
Output MIDI track number in `MidiFile.__iter__()`


page200:
When iterating through the events of each MIDI track separately, the track number is known, but the MIDI events are in general not in their temporal order in the case of a "MIDI Format 1" file.

`MidiFile.__iter__()` allows iterating through the events in their temporal order, and it would be nice to still receive information regarding which MIDI track each event came from. In the case of a "MIDI Format 1" file, different events might come from different MIDI tracks.

Just like `delta_seconds` (issue #527, PR #535) as well as the pitch bend in semitones and pitch-bend range (only General MIDI 1; issue #514, PR #535), the track number depends on earlier (meta-)events and shouldn't be additionally stored inside the current event. Thus, it can be stored together with `delta_seconds`, the pitch bend in semitones, and the pitch-bend range. In PR #535, those are stored in a "sidecar variable" that accompanies the MIDI event object. Alternatively, the MIDI event object can be nested into one larger object alongside all the other info.

rdoursenaud:
> When iterating through the events of each MIDI track separately, the track number is known, but the MIDI events are in general not in their temporal order in the case of a "MIDI Format 1" file.

That's an issue we need to take care of. This would be especially useful for format 2 files.

> On the other hand, `MidiFile.__iter__()` allows iterating through the events in their temporal order, but the MIDI track number associated with each event is not being provided to the user. Providing it to the user would be nice. In the case of a "MIDI Format 1" file, different events might come from different MIDI tracks.

Maybe, I'll have to think about it. Sounds reasonable enough but may lead to information duplication. Which then leads to potential inconsistencies and difficult resolution.

> Just like `delta_seconds` (issue #527, PR #535) as well as the pitch bend in semitones and pitch-bend range (only General MIDI 1; issue #514, PR #535), the track number depends on earlier (meta-)events and shouldn't be additionally stored inside the current event.

Agreed.

> Thus, it can be stored together with `delta_seconds`, the pitch bend in semitones, and the pitch-bend range. In PR #535, those are stored in a "sidecar variable" that accompanies the MIDI event object.

Still not convinced by this approach.

> Alternatively, the MIDI event object can be nested into one larger object alongside all the inferred info.

That's what a `Track` is, no? Let's make tracks more useful first, don't you think?

page200:
> > When iterating through the events of each MIDI track separately, the track number is known, but the MIDI events are in general not in their temporal order in the case of a "MIDI Format 1" file.
> 
[...]

PR #565, opened 2023-11-27
Defer computing merged_track

akx:
Follows up on #470.

There's no point in computing `merged_track` (quite expensively) until it's needed.

However, the comment https://github.com/mido/mido/pull/470#issuecomment-1705069143 should be addressed (separately): it's not clear that `merged_track` is a cached object, and changes to `tracks` or their contents will not be reflected automatically.

### DIFF
--- a/mido/midifiles/midifiles.py
+++ b/mido/midifiles/midifiles.py
@@ -307,6 +307,7 @@ class MidiFile:
         self.clip = clip
 
         self.tracks = []
+        self._merged_track = None
 
         if type not in range(3):
             raise ValueError(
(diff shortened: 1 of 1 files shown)


rdoursenaud:
Thanks, looks good on a surface level but…

Doesn't this defeat the purpose of pre-computing `merged_track` which was to avoid the delay before playback?

`merged_track` is still an utterly broken kludge anyway:
- it doesn't update when a track is modified and becomes out of sync with the object
- it seems to only be used for naive playback purposes where a better data structure or process would alleviate its need

I'm planning on getting rid of it for mido 2.x.

akx:
> Doesn't this defeat the purpose of pre-computing `merged_track` which was to avoid the delay before playback?

That moved the delay to _all_ instantiations of `MidiFile`, even if they didn't use `merged_track` (or iterate over the `MidiFile`), which is painful for downstream users (and energy bills thereof) of the library who don't need `merged_track` at all; see https://github.com/YatingMusic/miditoolkit/pull/44 that speeds up that library's test suite by quite a lot. (`check_msgdict` is still rather slow, but that's a different can of worms.)

rdoursenaud:
Yeah, it wasn't a smart decision on my part to accept this PR as-is.
I'm just trying to be cautious and aware of the repercussions.

rdoursenaud:
I believe your approach is the better tradeoff for now.
I'll have to think and find a better approach that fits both use cases for the next release.

akx:
@rdoursenaud I absolutely know the feeling of trying to be cautious (being the maintainer of [babel](https://github.com/python-babel/babel)) 😄 

As for 2.0, I think `__iter__` should probably be removed altogether (since #537 would necessitate a different API anyway); I think `MidiFile.iter_events(all_tracks=True)` sounds pretty good at first thought, etc... :)

rdoursenaud:
I'm new to the maintaining a popular library thing and still learning the ropes.
Thanks for your invaluable insights!
Feel free to stick around ;)

PR #584, opened 2024-04-22
Remove upper bound of `packaging`

carlthome:
`packaging` is now 24.0 so perhaps the upper version bound should be removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ classifiers = [
 
 dependencies = [
     "importlib_metadata; python_version<'3.8'",  # For version introspection
-    "packaging~=23.1",  # To provide a nice version_info opbject
+    "packaging>=23.1",  # To provide a nice version_info opbject
 ]
 dynamic = ["version"]
 


carlthome:
Would be nice to get this in! Any blockers to consider?

manoskary:
Requesting `packaging~=23.1` is creating a `ContextualVersionConflict` when installing Mido in Colab.
Although the default `packaging~=24` is uninstalled the kernel has to be restarted for this to be taken into effect.
This essentially makes the installation of Mido any packaging that includes Mido as a dependency not functional on the first try within Colab.

carlthome:
How about merging this unpinning @rdoursenaud, @olemb?

rdoursenaud:
Sorry for the long hiatus.
I’ve removed all version checking on hard dependencies to avoid any further complications.
I didn’t properly assess the impact of doing so previously and deeply apologize for the inconveniences created and the very long delay in properly addressing it.
Expect a hotfix release very soon.

PR #590, opened 2024-07-06
feat(RtMidi): quick implementation of timeout for receive

MrFrangipane:
Hi,

Thank you for this great library.

I'm developping a [project](https://github.com/MrFrangipane/osc-artnet-daemon/tree/rewrite) that uses MIDI receive in tight loops. In order to quit gracefully, I implemented for RtMidi backend a timeout argument.

I'll have to check deeper into the source of MIDI to complete this pull request.

Is this something useful for you ?

All the best


rdoursenaud:
Very sorry for the late reply, I've been quite overwhelmed with life. This may be useful to some users, yes. I personally vastly prefer using callbacks than polling but you seem to have a valid use case. Feel free to fledge it out!

PR #598, opened 2024-09-23
add --force-program option to mido-play

etene:
I use mido-play to pipe MIDI files into Fluidsynth. When loading a custom soundfont, I sometimes want the notes from a specific channel to be played with a specific instrument.
I can set that with `prog <channel> <number>` in Fluidsynth, however MIDI files often contain `program_change` events that will override my manual setting.

To work around this issue, I modified `mido-play` to be able to override the `program` value of `program_change` messages on a per-channel basis. I figured it could be useful to other users.

### DIFF
--- a/mido/scripts/mido_play.py
+++ b/mido/scripts/mido_play.py
@@ -16,10 +16,19 @@ Todo:
   - add option for printing messages
 """
 import argparse
+import re
 import sys
 
 import mido
 from mido import Message, MidiFile, tempo2bpm
+
+
+def program_override(value):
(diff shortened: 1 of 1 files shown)


rdoursenaud:
Thank you for your contribution. However, `mido-play` is just a sample program to demonstrate MIDO's capabilities. We don't intend to maintain it further and/or turn it into a fully-fledged MIDI player just yet. Thus I'm unfortunately unwilling to add any features to it. Feel free however to maintain your own fork for your purposes.

PR #628, opened 2025-07-18
Add github codespace support

oxygen-dioxide:
GitHub codespace is an online Linux environment with a vscode web UI. It is useful for developers who don't have a Linux device but want to test and debug their code on Linux.


rdoursenaud (changes requested):
Thanks for the suggestion. Please have a look at my comments.

PR #639, opened 2025-11-28
Remove extra comma in docs calling `from_hex`

nhallajian:
The `from_hex` code examples result in an error since they contain a comma so this PR removes them.


rdoursenaud (approved):
Nice catch! Thanks.

PR #648, opened 2026-04-13
Expose open and get functions without tricks

Leo-Tinkeam:
Remove the import that I have made last year because they do not respect DRY (Don't Repeat Yourself)

The new one are detected as functions instead of variables by type-checkers (better as they are functions)

Also added them to `__all__` to import them with `from mido import *`

Same for the 3 functions `open_xxx` that was not already exported statically (only dynamically)

I allowed the import of `backend` variable and `Backend` class to keep compatibility with programs that were using these

### DIFF
--- a/mido/__init__.py
+++ b/mido/__init__.py
@@ -93,10 +93,17 @@ Getting started:
     ['MPK mini MIDI 1', 'SH-201']
 """
 
-from typing import TYPE_CHECKING
-
 from . import ports, sockets
-from .backends.backend import Backend
+from .backends.backend import (
+    Backend,
+    get_input_names,
+    get_ioport_names,
+    get_output_names,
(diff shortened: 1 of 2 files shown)


Leo-Tinkeam:
If you need me to explain more in details for the review or to change something, just tell me

For the CI, as far as I understand, the problem was already there before

Leo-Tinkeam:
I'm not really aware about what milestones are. Should I do anything ?

If you want me to change something, feel free to ask

rdoursenaud:
No, you don't need to do anything.
Milestones help me track what changes I want to introduce in which version. I'd like to introduce yours in the next one but need a bit of time to review it properly, so I added it to the milestone to not forget about it when the time comes.
Thanks!

PR #650, opened 2026-06-25
Fix _parse_data: empty sysex str round-trip and malformed-paren detection

gaoflow:
## What

Two bugs in `mido/messages/strings.py::_parse_data()`:

### Bug 1 – Round-trip broken for empty sysex

`Message.from_str(str(Message('sysex', data=())))` raises `ValueError: unable to parse data bytes`.

When the data field is `()` (empty sysex), `"".split(',')` returns `['']` and `int('')` raises `ValueError`.  The fix adds an early return of `[]` when the inner content is empty:

```python
inner = value[1:-1]
if not inner:
    return []
```

### Bug 2 – Wrong boolean condition in the parenthesis guard

The original guard was:

```python
if not value.startswith('(') and value.endswith(')'):
```

This only raised `ValueError: missing parentheses in data message` when the *opening* `(` was absent **and** the closing `)` was present.  For any other malformed case — no parens at all, or a missing closing `)` — it fell through to a confusing "unable to parse data bytes" error.

Fixed to:

```python
if not (value.startswith('(') and value.endswith(')')):
```

## Reproducer

```python
from mido import Message

# Bug 1: round-trip
msg = Message('sysex', data=())
print(str(msg))          # 'sysex data=() time=0'
Message.from_str(str(msg))  # raised ValueError before this fix

# Bug 2: missing-paren error message
from mido.messages.strings import _parse_data
_parse_data('1,2,3')    # gave 'unable to parse data bytes', now gives 'missing parentheses'
_parse_data('(1,2,3')   # same wrong error; now correct
```

## Tests

Four new tests added to `tests/messages/test_strings.py`:
- `test_decode_empty_sysex` — decoding `data=()` returns empty tuple
- `test_sysex_str_roundtrip` — `from_str(str(msg)) == msg` for `()`, `(1,)`, `(1,2,3)`
- `test_decode_sysex_missing_parens` — no parens raises correct error
- `test_decode_sysex_missing_open_paren` — missing `(` raises correct error

All 126 tests pass.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/mido/messages/strings.py
+++ b/mido/messages/strings.py
@@ -40,11 +40,14 @@ def _parse_time(value):
 
 
 def _parse_data(value):
-    if not value.startswith('(') and value.endswith(')'):
+    if not (value.startswith('(') and value.endswith(')')):
         raise ValueError('missing parentheses in data message')
 
+    inner = value[1:-1]
+    if not inner:
+        return []
     try:
-        return [int(byte) for byte in value[1:-1].split(',')]
+        return [int(byte) for byte in inner.split(',')]
     except ValueError as ve:
         raise ValueError('unable to parse data bytes') from ve
 
--- a/tests/messages/test_strings.py
+++ b/tests/messages/test_strings.py
@@ -21,3 +21,23 @@ def test_encode_sysex():
     # This should not have an extra comma.
     assert str(Message('sysex', data=(1,))) == 'sysex data=(1) time=0'
     assert str(Message('sysex', data=(1, 2, 3))) == 'sysex data=(1,2,3) time=0'
+
+
+def test_decode_empty_sysex():
+    assert Message.from_str('sysex data=() time=0').data == ()
+
+
+def test_sysex_str_roundtrip():
+    for data in [(), (1,), (1, 2, 3)]:
+        msg = Message('sysex', data=data)
+        assert Message.from_str(str(msg)) == msg
+
+
+def test_decode_sysex_missing_parens():
+    with raises(ValueError, match='missing parentheses'):
+        Message.from_str('sysex data=1,2,3 time=0')
+
+
+def test_decode_sysex_missing_open_paren():
+    with raises(ValueError, match='missing parentheses'):
+        Message.from_str('sysex data=1,2,3) time=0')


mido/messages/strings.py
```diff
@@ -40,11 +40,14 @@ def _parse_time(value):
 
 
 def _parse_data(value):
-    if not value.startswith('(') and value.endswith(')'):
+    if not (value.startswith('(') and value.endswith(')')):
```
rdoursenaud, line 43: This does not make much more sense than the previous condition. From a cursory glance, it seems it should have beeen:
```py
if not value.startswith('(') or not value.endswith(')'):
    ...
```

mido/messages/strings.py
```diff
@@ -40,11 +40,14 @@ def _parse_time(value):
...
         raise ValueError('missing parentheses in data message')
 
+    inner = value[1:-1]
+    if not inner:
+        return []
     try:
-        return [int(byte) for byte in value[1:-1].split(',')]
```
rdoursenaud, line 47: I believe this could be simplified to:
```py
return [int(byte) for byte in value[1:-1].split(',') else None]
```

However, is a message without data even valid? I don't think so!

rdoursenaud (changes requested):

mido/messages/strings.py
rdoursenaud: Nevermind, I wasn't fully awake. The two statements are identical and produce the same thruth table. Yours reads better an has one less computation (if unoptimized by the JIT). Sorry for the noise!

mido/messages/strings.py
binggao1230: I treated the empty case as valid because Mido already accepts and emits it today:

```python
msg = Message('sysex', data=())
str(msg)  # 'sysex data=() time=0'
```

So this part of the PR is only making `_parse_data()` accept the string format that `Message.__str__()` already produces, preserving `Message.from_str(str(msg)) == msg`. If you prefer empty sysex to become invalid at the model level, that seems like a larger behavior change outside this parser round-trip fix.

binggao1230:
Thanks for the review. I replied inline on the empty sysex case: Mido already emits `sysex data=()` today via `str(Message('sysex', data=()))`, so this PR keeps `Message.from_str(str(msg)) == msg` for that existing representation rather than making empty sysex newly valid at the model level.

I do not think there is a code change needed after the clarified condition comment, but I am happy to adjust if you prefer treating empty sysex as invalid in a separate behavior change. Could you re-review when you have a chance?

rdoursenaud:
Yeah, you're right. Validation should only be done on I/O since manipulating partial objects is the whole point of MIDO. This is a major design change I'll tackle in version 2.0 once I'm able to get back to it. In the meantime, I'll approve your PR. Thank you very much!

rdoursenaud (approved):

PR #656, opened 2026-09-16
Clarify SysEx payload and message framing

### LINKED ISSUE #622: Clarify System Exclusive Messages documentation (Was: Encode / decode sysex bytes > 127 automatically)
When sending sysex messages, the user should be able to supply a list of
hex bytes without needing to encode values > 127.
What is the point of putting this burden on the user?

Please add the functionality for sending a sysex message as-is without
requiring the user to encode bytes > 127.

Thank you

xuu33030:
## Summary

Clarify the SysEx documentation as requested in #622:

- `data` contains only 7-bit payload bytes, not the `F0`/`F7` framing bytes that Mido adds when encoding or sending.
- Show `Message.from_bytes()` and `Message.from_hex()` for an already framed message, and link to the SysEx file documentation.
- Explain that any encoding of larger payload values is device-specific.

This preserves the existing validation and API behavior; it does not silently strip status bytes from `data`.

Closes #622.

## Validation

On Python 3.11.15, using the source checkout:

- Executed all 12 interactive examples in the updated SysEx section with Python's `doctest`: passed.
- Checked empty and boundary-value payload round trips and rejection of framing bytes / out-of-range values in `data`: passed.
- `python -m pytest`: 126 passed, both before and after the documentation edit.
- `sphinx-build -j auto -q -W -E --keep-going docs <output>`: passed before and after.
- `ruff check .` and `git diff --check`: passed.

No MIDI hardware transmission was tested. External-link, REUSE, and manifest checks were not run locally.

AI assistance was used to investigate, edit, and validate this change.


rdoursenaud (approved):

rdoursenaud:
Thank you, nice improvement!

